PR #7, opened 2016-02-29
Prevent ls from trying to ln a directory

cdosborn:


cdosborn:
@amercer1 this is untested

amercer1:
A few things. 
- [ ] Can you request that this gets pulled into our dev branch?
- [ ] I want everything except directories. 

If you can use your find command to also exclude directories that would be fantastic.

PR #9, opened 2016-02-29
Prevent ls from trying to ln a directory

cdosborn:


cdosborn:
Give this a test.

amercer1:
Alternative solution was found. Thanks!

PR #10, opened 2016-03-31
Remove rc.local.atmo

amercer1:
modified:   tasks/main.yml
	modified:   vars/CentOS-5.yml
	modified:   vars/CentOS-6.yml
	modified:   vars/CentOS-7.yml
	modified:   vars/Ubuntu-12.yml
	modified:   vars/Ubuntu.yml


nfaction:
We should have a variable list for the modprobe modules such that more can be added in the future.  This will also help if there is a module name discrepancy between different OS types and versions.

PR #23, opened 2016-04-25
Novnc

nfaction:
Replace RealVNC as option for Web-based VNC session.

Be sure to disable the playbook for RealVNC before using this role and playbook.


amercer1:
Added a util playbook and role to check for a valid novnc setup. Will work in conjunction with troposphere and its ability to render a `web desktop` button.

amercer1:
💃

PR #25, opened 2016-05-12
Nested playbook reorganization

steve-gregory:
Heres an example of how we could reorganize the playbooks to condense the number being executed and try to make sense of the dependencies between playbooks.

For our solution, we are already giving atmosphere a PLAYBOOK_DIR setting, by using a nested structure rather than parallel, we don't have to "dive a directory back" to select `util_playbooks` or other `name_name_playbooks`.

new values look like this in [The atmosphere PR](https://github.com/iPlantCollaborativeOpenSource/atmosphere/pull/144):
```
#Old method -- more prone to failure:
playbooks_dir = settings.ANSIBLE_PLAYBOOKS_DIR
deploy_playbooks = playbooks_dir
util_playbooks = playbooks_dir.replace('/playbooks', '/util_playbooks')

#New Method:
playbooks_dir = settings.ANSIBLE_PLAYBOOKS_DIR
deploy_playbooks = os.path.join(playbooks_dir, 'instance_deploy')
util_playbooks = os.path.join(playbooks_dir, 'utils')
```

The current implementation keeps the order of execution *identical* to that found in main stream master, the only difference is in how the plays are grouped.


ansible/roles/atmo-user-ssh-keys/tasks/main.yml
```diff
@@ -8,24 +8,3 @@
...
-    - "{{ ATMOUSERNAME }}"
-  tags:
-    - ssh-keys
-    - debug
-
-- name: remove user key from controller node
-  local_action: file path=/tmp/{{ ATMOUSERNAME }}_default.pub state=absent
-  tags:
-    - ssh-keys
-    - delete-key
```
nfaction: This code should not be removed as this is required for passwordless login for GateOne.  Really this should be split out into different roles.

ansible/roles/gateone-gen-sshkey/tasks/main.yml
```diff
@@ -30,5 +29,30 @@
   delegate_to: gateone
 
 - name: finally, transfer the public key locally
-  fetch: src="/var/lib/gateone/users/{{ ATMOUSERNAME }}/.ssh/{{ ATMOUSERNAME }}_default.pub" dest="{{ GATEONE_LOCAL_PUBKEY_DIR | default('/tmp/') }}" flat=yes
+  fetch: src= "/var/lib/gateone/users/{{ ATMOUSERNAME }}/.ssh/{{ ATMOUSERNAME }}_default.pub" dest= "{{ GATEONE_LOCAL_PUBKEY_DIR | default('/tmp/') }}" flat=yes
```
nfaction: This might break code.  Likely a syntax error.  All should be checked with a `--syntax-check`

ansible/roles/gateone-gen-sshkey/tasks/main.yml
```diff
@@ -30,5 +29,30 @@
...
+#     - "{{ ATMOUSERNAME }}"
+#   tags:
+#     - ssh-keys
+#     - debug
+# 
+# - name: remove user key from controller node
+#   local_action: file path=/tmp/{{ ATMOUSERNAME }}_default.pub state=absent
+#   tags:
+#     - ssh-keys
+#     - delete-key
```
nfaction: This should be re-enabled

ansible/roles/gateone-gen-sshkey/tasks/main.yml
steve-gregory: This line executes fine//is syntactically correct.. I added the space was added because `vim`'s parser goes ape- ⁉️ when you `="`

The other option would be to nest the values instead of inline?

ansible/roles/atmo-user-ssh-keys/tasks/main.yml
steve-gregory: Yep, I was moving this code into `gateone-` as you noted below.. but..

ansible/roles/gateone-gen-sshkey/tasks/main.yml
steve-gregory: As noted [above on L35](https://github.com/iPlantCollaborativeOpenSource/atmosphere-ansible/pull/25/files#diff-8a0ac20f8fd8e19efb9f578e63c14697R35) I commented it out because I was receiving this error:
```
fatal: [129.114.5.19]: FAILED! => {"failed": true, "msg": "the file_name '/tmp/sgregory_default.pub' does not exist, or is not readable"}
```

amercer1: @steve-gregory You got that error because you didn't run the other relevant gateone playbook first

nfaction: This role expects the previous role to be run before-hand, since that role is responsible for generating and depositing keys on the instance for GateOne.  To fix this, please run `gateone-gen-ssh-key`.  This is the section of code that is responsible for that file: https://github.com/iPlantCollaborativeOpenSource/atmosphere-ansible/blob/master/ansible/roles/gateone-gen-sshkey/tasks/main.yml#L32-L34

ansible/playbooks/instance_deploy/30_post_user_install.yml
```diff
@@ -0,0 +1,9 @@
+---
+# Requires ATMOUSERNAME and VNCLICENSE
+- name: Playbook lists roles that should be executed after ATMOUSERNAME has been created
+  hosts: atmosphere
+  roles:
+    - atmo-fail2ban
+    - atmo-irods
+    - atmo-realvncserver
+    - atmo-cleanup
```
nfaction: Add `  - gateone-gen-sshkey` after `atmo-realvncserver` in this playbook

Issue #30, opened 2016-06-14
CentOS 5 systems time set in future prevent yum commands


nfaction:
<http://superuser.com/questions/635020/how-to-know-current-time-from-internet-from-command-line-in-linux>

```
-bash-3.2# hwclock -s
-bash-3.2# date
Tue Jun 14 22:48:59 MST 2016
-bash-3.2# date -s "14 JUN 2016 15:48:00"
Tue Jun 14 15:48:00 MST 2016

OR

date -s "$(curl -s --head http://google.com | grep ^Date: | sed 's/Date: //g')"
```

The time update MUST OCCUR BEFORE ntp package install for CentOS systems

nfaction:
Fixed and merged from branch `instance-time-issue`

PR #31, opened 2016-06-15
Add task to manually set time with fallback

nfaction:
An error occurred when yum tries to install ntp packages, since the instances time was too far in the future for it to work.  Now time is being set with the hardware clock and a remote time query.

	modified:   ansible/roles/atmo-ntp/tasks/main.yml


nfaction:
This PR is a fix for issue #30 

Please review @amercer1

amercer1:
🚢 

💃

PR #33, opened 2016-06-20
Sshd config dupe port

calvinmclean:
Modified atmo-ssh-setup role so that the only port entry in `sshd_config` will be 'Port 22'
- First removes lines with `regexp` '^#Port' or '^Port' for a clean slate
- Uses `regexp` and `backrefs` to find the line about port entries and insert 'Port 22' on the next
  - the `regexp` is dependent on the OS. Included tasks for Ubuntu and CentOS
- Only makes the change if the `regexp` is present (vs `insertafter` which will always make a change)


nfaction:
@calvinmclean Could you please launch a set of test instances on all of our clouds and manually verify that it does what it is supposed to do?  It does look good, but before we merge, I would like direct confirmation on testing this.  We can work together on this if you like.

calvinmclean:
I did more testing this morning. I launched an Ubuntu and CentOS system on each cloud (Tucson, Tucson Workshop, and Austin) and ran a play with just the tasks that I added and each one worked as expected.

steve-gregory:
@nfaction Is this ready :shipit: ?

steve-gregory:
@calvinmclean can you fix the merge conflicts that exist in this PR, then we will :shipit:

calvinmclean:
@steve-gregory I'll get started with that. How can I see where the conflicts are? There were none last time I looked at this.

ansible/roles/atmo-ssh-setup/tasks/main.yml
```diff
@@ -102,6 +102,23 @@
...
+    - { regexp: '^PermitRootLogin', line: 'PermitRootLogin without-password'}
+    - { regexp: '^UseDNS', line: 'UseDNS no'}
+
+  #if port line is commented or not commented out it is removed
+- name: ensure that the ports are correct in sshd_config
+  lineinfile: dest=/etc/ssh/sshd_config backup=yes regexp="^Port" state=absent
+- lineinfile: dest=/etc/ssh/sshd_config backup=yes regexp="^#Port" state=absent
+
+  #add 'Port 22' entry
+- lineinfile: dest=/etc/ssh/sshd_config backup=yes backrefs=yes regexp="^# What ports, IPs and protocols we listen for" line="# What ports, IPs and protocols we listen for\nPort 22"
```
amercer1: Should these `Port 22` lines be hard coded?

PR #34, opened 2016-06-24
Add task to remove unwanted or no longer used ssh keys

nfaction:
Add task to remove SSH public key early on in the deployment process.  Fix syntax to prevent deprecation warnings and handle variables that are not already defined.

This requires the NEW variable `SSH_KEYS_TO_REMOVE` in `group_vars/all` in order to remove keys.

	modified:   roles/atmo-ssh-setup/defaults/main.yml
	modified:   roles/atmo-ssh-setup/tasks/setup-root.yml


ansible/roles/atmo-ssh-setup/tasks/setup-root.yml
```diff
@@ -10,9 +10,16 @@
...
   remote_user: "{{ use_remote_user }}"
 
-- name: add ssh keys to authorzied_keys
+- name: add ssh keys to authorized_keys
   authorized_key: user=root key="{{ item }}" state=present
   become: true
   remote_user: "{{ use_remote_user }}"
-  with_items:
-    SSHKEYS
+  with_items: '{{ SSHKEYS }}'
```
amercer1: Yay for 2.x syntax.

amercer1:
🚢

PR #40, opened 2016-07-18
Prevent modules failure from throwing total failures

amercer1:


cdosborn:
how can we review if you merge so quickly 😢

PR #46, opened 2016-08-17
Add atmo-make-space-on-instace role

amercer1:
This role is to remove files that are taken up too much space on already filled instances.

	new file:   playbooks/instance_deploy/05_make_space_on_instace.yml
	new file:   roles/atmo-make-space-on-instace/defaults/main.yml
	new file:   roles/atmo-make-space-on-instace/tasks/main.yml


steve-gregory:
Code LG2M  👍 

Requesting a rename to `atmo-pre-clean` rather than `atmo-make-space-on-instance` before merging.

amercer1:
💃

cdosborn:
@amercer1 and I had an offline chat.
In summary this needs to be reverted for the following reasons:

- The conditional (90%) is bad 
    - this will create hard to debug bugs 
    - "Why is stuff being deleted only part of the time?"
- It's bad that we are deleting user's files

When we encounter an issue (a user running out of disk space). We need to be more skeptical of whether it deserves a fix, whether it is worth adding complexity. Here the proper design is to let the user get burned by their own actions. Always deleting files **we** download is a possible way forward.

PR #47, opened 2016-08-18
Revert "Add atmo-make-space-on-instace role"

amercer1:
Pulling this out as it may be too invasive on user's files. Trying again.


cdosborn:
👍

PR #53, opened 2016-10-07
Check novnc proxy network

amercer1:
This additional task is to ensure that then novnc proxy can talk to the newly launched instance. Although this is a minor check that is performed only at deployment/redeployment, it will help us prevent rendering web_desktop when a connection can't be made.


steve-gregory:
:shipit: thank you @amercer1

julianpistorius:
Nice @amercer1! 👍

lenards:
I'm super happy to see this make it in!

up-kid-wow

PR #55, opened 2016-10-28
Print all known facts/variables for given inventory_hostname

steve-gregory:
New utility playbook will print variables for an instance.


steve-gregory:
:shipit:

PR #56, opened 2016-11-02
Run distro updates

c-mart:
This implements the change proposed in JIRA ATMO-1578. Text of JIRA description follows.

> Edwin and Steve observed that https://atmo.cyverse.org/application/images/1009 fails deployment because of an APT error when installing NTP.

> This is an Ubuntu 12.04 image that hasn't had distro/security updates in over three years, so we're trying to install modern/patched NTP alongside an ancient OpenSSL dependency. There's a launchpad bug (https://bugs.launchpad.net/ubuntu/+source/ntp/+bug/1630877), and the solution (to this and many other problems!) is just to run distro updates. I tested and confirmed the fix.

> This raises the question of why we're offering instances that are so far behind on updates. I propose that atmosphere-ansible should just run distro updates early in the deployment process for all instances.

> Benefits:
> - Huge security win for Atmosphere platform, users will get instances with an up-to-date kernel and packages
> - Reduces some maintenance workload from image maintainers, who generally won't need to worry about regularly applying updates to images (as long as updates are still provided by the distro maintainers)
> - Easy to implement, atmosphere-ansible can just consume my https://github.com/CyVerse-Ansible/ansible-distro-update from Ansible Galaxy
> - Fixes the NTP issue

> Risks:
> - Instance deployment will take a little longer. The older the image, the longer updates will take. We could minimize this by updating images more frequently, at least the base images, like Jetstream does.
> - It's possible for an image maintainer to configure their image in such a way that running updates breaks functionality in some way. We would need to decide how to handle that if it happens. In whatever contract / distribution of responsibilities we have with image maintainers, we could advise them that we will run updates when instances are deployed.

Before we deploy this to production, we should test and verify that it does not break anything related to deployment of our most important/popular images. I have already confirmed that it works for the Ubuntu 12.04 image that originally prompted this.


c-mart:
Closing this so I can free up my fork's master branch, will open a new PR from a feature branch.

PR #59, opened 2017-01-20
Revert "Merge pull request #33 from calvinmclean/sshd_config-dupe-port"

cdosborn:
This reverts commit 03b89954ebe316fc3c390e0e92f31382f8c2cefb, reversing
changes made to 0b1939f15c6d7a720e06dc280c27535f9d85ae5e.

### DIFF
--- a/ansible/roles/atmo-ssh-setup/defaults/main.yml
+++ b/ansible/roles/atmo-ssh-setup/defaults/main.yml
@@ -4,4 +4,3 @@ SSH_OPTIONS_TO_PASS_IN: "-o PreferredAuthentications=publickey"
 
 SSH_ALLOW_GROUPS: "root users"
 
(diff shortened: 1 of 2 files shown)


c-mart:
For the record, Connor found that the previous merge broke instance deployments, because fact gathering was not performed on the target before conditional statements like `when: ansible_distribution == "CentOS"`, which causes a failure due to undefined variables. I suggested that he revert the merge because the broken atmosphere-ansible was preventing him from working on a different problem. Subsequently (but not today), we plan to refactor the `atmo-ssh-setup` playbook to resolve issues that the original pull request was trying to address.

c-mart:
Merging this as master is currently broken and I need to test some things which depend on atmosphere-ansible. :)

Issue #29, opened 2016-06-14
Remove duplicate `Port` entry for `sshd_config`


nfaction:
Ubuntu 14.04.2 XFCE Base instance contains both:

```
Port 1657
Port 22
```

Should add either lineinfile to remove all entries and add just the desired port for SSH using regex to remove, then add again.  Only then SSH can be restarted

calvinmclean:
@c-mart @steve-gregory 
I believe we can close this issue after PR #62

calvinmclean:
Actually, I just launched an instance and both port entries are still there...

lenards:
Fixed by #66.

PR #62, opened 2017-02-01
Rework atmo-ssh-setup role

calvinmclean:
This role had some weird things going on that turned out to be necessary. The changes I made mostly make these things more straightforward and easy to understand, as well as updated the syntax and styling.

Since facts are not gathered for this role, the first tasks use `local_action` to attempt to connect with SSH and determine the correct username to use in this instance. I separated these tasks in `main.yml` and put "regular" tasks into `ssh-setup.yml` which is included with `become` so that it uses `root` user. Also in `main.yml` is a setup task for Ubuntu 16.04, since Ansible 2.1 requires Python 2.7.

Tested on Ubuntu 12.04, Ubuntu 14.04, CentOS 5, and CentOS 6 on `atmobeta`
Tested on Ubuntu 12.04, Ubuntu 14.04, Ubuntu 16.04, CentOS 5, CentOS 6, and CentOS 7 with vagrant

quick list:
1. Fixes duplicate port entries (issue #29)
2. Reorganized
3. Fix bad practices and simplified many tasks with lists and vars


ansible/roles/atmo-ssh-setup/tasks/main.yml
```diff
@@ -1,145 +1,73 @@
...
+    set_fact: use_remote_user=ubuntu
+    when: not ubuntu_connection|skipped and ubuntu_connection.rc == 0
+
+  - name: Set use_remote_user variable to "root"
+    set_fact: use_remote_user=root
+    when: root_connection.rc == 0
+  tags: set_facts
+
+# Install Python2.7 for Ubuntu 16.04
+- block:
```
c-mart: This is a lot more sophisticated than the way I was doing it!

ansible/roles/atmo-ssh-setup/tasks/root-setup.yml
```diff
@@ -0,0 +1,26 @@
+---
+
+- name: Clear the tty from suoers
+  raw: sed -i '/^Defaults\s*requiretty/d' /etc/sudoers
+
+- name: Remove the standard cloud disable command
```
c-mart: minor nitpick, I don't think this task needs to be run when `use_remote_user` is root (i.e. when root SSH login already works). You could add something like `when: 'use_remote_user != "root"'` to effect this

ansible/roles/atmo-ssh-setup/tasks/ssh-setup.yml
```diff
@@ -0,0 +1,42 @@
+---
+
+- name: Directly check linux distribution
```
c-mart: This task is a little wonky (on Ubuntu that file won't be there) but I understand what you're doing.

An alternative approach is running the [setup](http://docs.ansible.com/ansible/setup_module.html) module (which discovers the actual distro), and then gather OS-specific variables [this way](https://github.com/c-mart/atmosphere-ansible/blob/c10837781229281f11606a5d859c6e3ad7f41936/ansible/roles/atmo-dhcp/tasks/main.yml#L8). (At this point in the role you should be able to gather facts, as you already have a working Python 2 interpreter and you know which user you can log in as.)

ansible/roles/atmo-ssh-setup/tasks/ssh-setup.yml
```diff
@@ -0,0 +1,42 @@
...
+    backup: true
+    regexp: "{{ item.regexp }}"
+    line: "{{ item.line }}"
+    state: "{{ item.state }}"
+  with_items:
+    - "{{ SSHD_CHANGES }}"
+    - { regexp: "{{ PORT_ENTRY_REGEXP }}", line: "{{ PORT_ENTRY_LINE }}", state: 'present' }
+  tags: sshd-config
+
+- name: Append AllowGroups Line to /etc/ssh/sshd_config
```
c-mart: Why is this task separate from the task above? Could it just be another item in the `with_items` loop?

calvinmclean: @c-mart I kept this in a separate task because it does not use a `regexp`. But I suppose just setting `regexp: ""` might work since `lineinfile` defaults to EOF so I will give that a try.

ansible/roles/atmo-ssh-setup/tasks/ssh-setup.yml
calvinmclean: Good point, I will use setup instead.

Also, I didn't set the python interpreter for centos5 bc it didn't seem to be doing anything since `/usr/bin/python` is already in use. I didn't run into any problems because of this in vagrant VM or in my atmobeta testing.

calvinmclean:
Added a few changes related to the comments.

ansible/roles/atmo-ssh-setup/tasks/main.yml
```diff
@@ -1,145 +1,73 @@
...
+    when: not centos_connection|skipped and centos_connection.rc == 0
+
+  - name: Set use_remote_user variable to "ubuntu"
+    set_fact: use_remote_user=ubuntu
+    when: not ubuntu_connection|skipped and ubuntu_connection.rc == 0
+
+  - name: Set use_remote_user variable to "root"
+    set_fact: use_remote_user=root
+    when: root_connection.rc == 0
+  tags: set_facts
```
cdosborn: The rest of our playbooks depend on `use_remote_user` being defined. This block can leave the variable undefined. The last item in this block should include:
```
- fail:
    msg: "No remote connection established, atmo-ssh-setup can not continue."
  when: use_remote_user is not defined
```

ansible/roles/atmo-ssh-setup/tasks/ssh-setup.yml
```diff
@@ -0,0 +1,42 @@
...
+  lineinfile:
+    dest: /etc/ssh/sshd_config
+    backup: true
+    regexp: "{{ item.regexp }}"
+    line: "{{ item.line }}"
+    state: "{{ item.state }}"
+  with_items:
+    - "{{ SSHD_CHANGES }}"
+    - { regexp: "{{ PORT_ENTRY_REGEXP }}", line: "{{ PORT_ENTRY_LINE }}", state: 'present' }
+  tags: sshd-config
```
cdosborn: I think this is the exact use case for the `template` module. We could template sshd_config file. We wouldn't have to think about which lines to remove (rn we're removing `ALLOW_USERS` etc). I think we could join these sshd_config tasks into a single task too.

ansible/roles/atmo-ssh-setup/vars/CentOS.yml
```diff
@@ -0,0 +1,7 @@
+---
+
+SSH_SERVICE: sshd
+
+PORT_ENTRY_REGEXP: "^# default value."
+
+PORT_ENTRY_LINE: "# default value.\nPort {{ SSH_PORT }}"
```
calvinmclean: My reasoning for this is that sometimes the "Port" line is commented out, so I had lineinfile to remove the port line if it is commented or uncommented and then add it again later. The reason for matching the weird comments is so that the Port entry is in the same place it would regularly be in.

However, after looking into it some more, it doesnt look like the Port line is often commented out by default so I can probably drop the weirdness. If the "Port" line is not found, is it fine to just append the Port entry to the end of file?

calvinmclean:
Implemented changes relating to Connor's suggestions:
- Remove tags
- Fail when use_remote_user is not defined after running block
- Better grep
- Ubuntu 16.04 Python comments
- Remove unneeded conditional check
- Changed `Port` entry regex to be more straightforward
- Consolidate the two `AllowGroups` lineinfiles

Also, I agree that using an `sshd_config` template could be useful.

ansible/roles/atmo-ssh-setup/tasks/main.yml
```diff
@@ -1,145 +1,71 @@
...
+    when: not ubuntu_connection|skipped and ubuntu_connection.rc == 0
+
+  - name: Set use_remote_user variable to "root"
+    set_fact: use_remote_user=root
+    when: root_connection.rc == 0
+
+# Install Python2.7 for Ubuntu 16.04 since it uses Python3 by default and ansible_distribution
+# requires Python2.7. Ansible 2.2 includes experimental support for Python3
+- block:
+  - name: Manual Ubuntu version detection
```
cdosborn: Small refactor, move this task outside the block, then add `when: use_remote_user == "ubuntu" && ubuntu16.rc == 0` as a clause to the block not each individual task. I'm not sure if the first part of the when is necessary.

cdosborn:
:+1: Looks great

c-mart:
@calvinmclean I will merge once you confirm you have re-tested your most recent changes.

calvinmclean:
@c-mart Everything is tested on atmobeta except for the last commit that just removed extra white space, which makes no difference.

PR #63, opened 2017-02-06
Allowing deployers to prevent atmo-ansible from setting default NTP servers

c-mart:
@steve-gregory one more tweak incoming


c-mart:
Ack, requested PR against wrong base, abort!

PR #66, opened 2017-02-10
Remove duplicate port entries in sshd_config

calvinmclean:
These changes will remove any non-22 port entry and will uncomment the default entry if it is commented. This is a fix for issue #29 

I added a new task for this lineinfile because I did not want to add `backrefs: false` to each dictionary in the list.

There are other ways to do this. If this isn't a preferred solution I can offer up another.

Also, since these changes are for a problem that occurs in specific images that have weird default `sshd_config` I believe it creates a good argument for a `sshd_config` template for more consistency.

### DIFF
--- a/ansible/roles/atmo-ssh-setup/defaults/main.yml
+++ b/ansible/roles/atmo-ssh-setup/defaults/main.yml
@@ -19,10 +19,14 @@ SSHD_CHANGES:
     line: 'PasswordAuthentication yes'
     state: 'present'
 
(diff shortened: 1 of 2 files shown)


c-mart:
Suggest that Calvin take a different approach (because I don't understand the logic very well and see some corner cases where it may fail):
- First, remove any/all "port" directives from the file, whether commented out or not.
- Seconfd, insert only the port directive that we want (defined in variables)

c-mart:
merged!

PR #67, opened 2017-02-14
Avoid ansible authorized_key module error when user directory is missing

steve-gregory:
Ansible would bubble up an error: `getpwnam(): name not found`
This issue was documented in ansible, but unfortunately looks like its been passed between repos and is no longer being actively monitored (Based on the commit messages and dates, it looks like `ansible/ansible` is the best place for getting a _real fix_ to the ansible module issue):
https://github.com/ansible/ansible/issues/11257
https://github.com/ansible/ansible-modules-core/issues/1515

In our case, we can avoid potential errors by looking up the users home directory before we attempt to use the `authorized_key` module.


ansible/roles/atmo-user-ssh-keys/tasks/main.yml
```diff
@@ -1,11 +1,26 @@
...
   with_items: USERSSHKEYS
   become: yes
   when: USERSSHKEYS is defined
 
+# See https://github.com/ansible/ansible/issues/11257 , https://github.com/ansible/ansible-modules-core/issues/1515
+- name: verify that home directory of user exists (to avoid ansible error)
+  stat:
+    path: "/home/{{ ATMOUSERNAME }}"
+  register: user_home_dir
+  when: ATMOUSERNAME is defined
```
cdosborn: Why doesn't the user have a home directory? Is there a task for that?

steve-gregory: The user _should_ have a home directory by the time this playbook executes, in a normal execution flow.

 I encountered this in a very specific corner case and the solution above got me through.

PR #69, opened 2017-02-17
Change desktop background image on GUI machines

calvinmclean:
`ATMO-1643` - Added tasks to set new CyVerse wallpaper

Tested by launching featured images on atmobeta. All GUI instances had the new wallpaper and all non-GUI instances deployed normally.


c-mart:
Does this also work for images with an xfce desktop environment?

calvinmclean:
@c-mart Yes, I tested it on Ubuntu XFCE and CentOS with Gnome. Let me know if there are other tests I should do.

c-mart:
Cool. My only other minor comment is that you put this is the atmo-realvncserver role. While it certainly works for now, in the future we could want to support other protocols than VNC, in which case this should become its own role.

steve-gregory:
@calvinmclean looks good! I've added a few TODO items to the PR description. Thanks! 👍

PR #72, opened 2017-02-22
Fail2ban always started

c-mart:
Depends on #71. CentOS images were deploying with fail2ban stopped. Apparently one of the other tasks in the atmo-fail2ban role stopped the service, and if we don't need to template out any changes to the jail config files, the service was never started again. This ensures fail2ban is started when we are done configuring it.

### DIFF
--- a/ansible/roles/atmo-common/tasks/main.yml
+++ b/ansible/roles/atmo-common/tasks/main.yml
@@ -59,6 +59,11 @@
   tags:
     - timezone
 
+- name: Restart rsyslog so that logs are written with correct timezone
+  service:
+    name: rsyslog
+    state: restarted
+
 ########
 # MOTD #
 ########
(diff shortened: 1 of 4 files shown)


c-mart:
I was not completely correct - later in the `atmo-cleanup` role we also uninstall firewalld and this action apparently stops fail2ban. Added a commit to ensure it's started again afterward.

ansible/roles/atmo-common/tasks/main.yml
```diff
@@ -59,6 +59,11 @@
   tags:
     - timezone
 
+- name: Restart rsyslog so that logs are written with correct timezone
+  service:
+    name: rsyslog
+    state: restarted
```
steve-gregory: Looks like the same from #71 ?

c-mart: Yep, this depends on #71, just merge them in order.

PR #73, opened 2017-02-22
Updating irods-icommands role to 4.1.9-cv (CyVerse-optimized)

c-mart:
See [ansible-irods-icommands #1](https://github.com/CyVerse-Ansible/ansible-irods-icommands/pull/1), tested and confident it works. These new iCommands packages use Ilyoung Choi's [recent performance tweaks](https://github.com/cyverse/cyverse-irods/commit/43835120a33a50f57eccb6b06fe5a829949bcd90).

### DIFF
--- a/ansible/requirements.yml
+++ b/ansible/requirements.yml
@@ -1,5 +1,8 @@
 ---
 
-# from Galaxy
+# Required roles from Ansible Galaxy
+# Install latest version of required Ansible roles by running the following:
+# ansible-galaxy install -f -r requirements.yml -p roles/
+
 - src: CyVerse-Ansible.irods-icommands
   name: irods-icommands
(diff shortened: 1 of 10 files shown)


c-mart:
I also tested running this role over the top of its previous version, e.g. what would happen when redeploying an existing instance with the non-optimized iCommands/FUSE client. Since the .deb and .rpm package versions are the same, the package manager thinks the package is already installed, and doesn't make any changes.

If we want to get around this (e.g. if we want to support deploying the updated client to existing instances when a user clicks redeploy), we could do things like uninstall and then reinstall the package, but this complicates idempotence. I'd be keen to avoid that and just ask users launch a new instance, unless the "redeploy new client to existing instance" use case is an important goal.

c-mart:
Self-:shipit: as I'm confident it works and stakeholders are eager to get this deployed to production.

PR #74, opened 2017-02-23
Fix syslog service name on CentOS 5

c-mart:
I didn't bother testing #71 on CentOS 5 given its demise in a month. Sure enough, the next day someone reported that the fix breaks atmosphere-ansible when run against CentOS 5. CentOS 5 uses syslog, not rsyslog.


c-mart:
Tested and working with CentOS 5, CentOS 6, Ubuntu 14.04 -- confident this is good. :shipit:

cdosborn:
Thanks @c-mart

PR #76, opened 2017-03-01
Fix bug in atmo-ssh-setup

calvinmclean:
@c-mart pointed out that the tasks used to determine the connecting user would have a return code of 0 even when the connection fails.

This fix goes back to the old method of echoing a string, but does not grep for it and just checks stdout for the string in future conditionals.


c-mart:
Calvin assured me he tested on atmobeta using a "base" Ubuntu 14.04 cloud image, and our featured images.

PR #77, opened 2017-03-21
Problem: NTP_PREP_PACKAGE not installed

steve-gregory:
Solution: Use 'shell' instead of 'action'+'apt' for installing libssl1.0.0 (the NTP_PREP_PACKAGE)

### DIFF
--- a/ansible/roles/atmo-ntp/tasks/main.yml
+++ b/ansible/roles/atmo-ntp/tasks/main.yml
@@ -32,12 +32,19 @@
   tags:
     - time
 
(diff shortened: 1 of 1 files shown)


steve-gregory:
@c-mart i wish i could give you more background. In fact shell/command *does* work and *apt* module does not.

If you find the solution to this, we would gladly accept a PR to replace this code with working `apt` module!

c-mart:
In that case, if `command` works then why would we use `shell`?

PR #79, opened 2017-03-27
Improved volume creation support in Atmosphere

steve-gregory:
By using ansible modules `filesystem:` and `mount:`

# Related PRs
https://github.com/cyverse/atmosphere/pull/300

### DIFF
--- a/ansible/playbooks/utils/check_volume.yml
+++ b/ansible/playbooks/utils/check_volume.yml
@@ -9,10 +9,10 @@
   tasks:
 
     - fail: msg="Variable VOLUME_DEVICE is not set."
-      when: "{{ VOLUME_DEVICE is not defined }}"
+      when: VOLUME_DEVICE is not defined
 
     - fail: msg="Variable VOLUME_DEVICE_TYPE is not set."
-      when: "{{ VOLUME_DEVICE_TYPE is not defined }}"
+      when: VOLUME_DEVICE_TYPE is not defined
 
 
 - name: Play to ensure filesystem is present before mounting
--- a/ansible/playbooks/utils/check_volume.yml
+++ b/ansible/playbooks/utils/check_volume.yml
@@ -4,12 +4,9 @@
   hosts: all
   vars:
     example:
-        - "-e \"{'VOLUME_DEVICE':'/dev/sdb', 'VOLUME_MOUNT_LOCATION':'/vol-b', 'VOLUME_DEVICE_TYPE': 'ext4'}\""
+      - "-e \"{'VOLUME_DEVICE':'/dev/sdb', 'VOLUME_DEVICE_TYPE': 'ext4'}\""
 
   tasks:
-
-    - fail: msg="Variable VOLUME_MOUNT_LOCATION is not set."
-      when: "{{ VOLUME_MOUNT_LOCATION is not defined }}"
 
     - fail: msg="Variable VOLUME_DEVICE is not set."
       when: "{{ VOLUME_DEVICE is not defined }}"
@@ -18,7 +15,7 @@
       when: "{{ VOLUME_DEVICE_TYPE is not defined }}"
 
 
-- name: Play to mount a volume
+- name: Play to ensure filesystem is present before mounting
   hosts: atmosphere
   roles:
-    - atmo-mount-volume
+    - atmo-mkfs-volume
--- a/ansible/playbooks/utils/mount_volume.yml
+++ b/ansible/playbooks/utils/mount_volume.yml
@@ -9,13 +9,13 @@
   tasks:
 
     - fail: msg="Variable VOLUME_MOUNT_LOCATION is not set."
-      when: "{{ VOLUME_MOUNT_LOCATION is not defined }}"
+      when: VOLUME_MOUNT_LOCATION is not defined
 
     - fail: msg="Variable VOLUME_DEVICE is not set."
-      when: "{{ VOLUME_DEVICE is not defined }}"
+      when: VOLUME_DEVICE is not defined
 
     - fail: msg="Variable VOLUME_DEVICE_TYPE is not set."
-      when: "{{ VOLUME_DEVICE_TYPE is not defined }}"
+      when: VOLUME_DEVICE_TYPE is not defined
 
 
 - name: Play to mount a volume
--- a/ansible/playbooks/utils/unmount_volume.yml
+++ b/ansible/playbooks/utils/unmount_volume.yml
@@ -8,7 +8,7 @@
   tasks:
 
     - fail: msg=" Must set pg_version. Pass a value in via `-e` as JSON {{example[0]}}"
-      when: "{{ src is undefined }}"
+      when: "{{ src is not defined }}"
 
 - name: Playbook to unmount a volume
   hosts: atmosphere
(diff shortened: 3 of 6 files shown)


ansible/playbooks/utils/check_volume.yml
```diff
@@ -0,0 +1,21 @@
...
+
+- name: Checking for necessary --extra_vars
+  hosts: all
+  vars:
+    example:
+      - "-e \"{'VOLUME_DEVICE':'/vol1', 'VOLUME_MKFS_DEVICE':True, 'VOLUME_DEVICE_TYPE': 'ext4'}\""
+
+  tasks:
+
+    - fail: msg="Must set VOLUME_DEVICE. Pass a value in via `-e` as JSON {{example[0]}}"
```
c-mart: Ansible doesn't care how a variable is passed (whether via extra_vars or any other way), and besides, I don't think that is JSON, it's YAML ["inline hash" format](http://yaml.org/YAML_for_ruby.html#simple_inline_hash). So, I would just say "Must set VOLUME_DEVICE". (Same applies everywhere else in this PR that you mention JSON).

Also, why do you make `example` a list and then index into its first element, when it could just be a string?

ansible/playbooks/utils/check_volume.yml
```diff
@@ -0,0 +1,21 @@
...
+      when: "{{ VOLUME_DEVICE is undefined }}"
```
c-mart: Does this code do what you intend? I think the double curly braces are for variable expansion, and I don't think `is undefined` means anything.

I think we want `when: VOLUME_DEVICE is not defined`

ansible/playbooks/utils/check_volume.yml
```diff
@@ -0,0 +1,21 @@
...
+
+    - fail: msg="Must set VOLUME_DEVICE_TYPE. Pass a value in via `-e` as JSON {{example[0]}}"
+      when: "{{ VOLUME_DEVICE_TYPE is undefined }}"
+
+
+- name: Playbook to test volume before mounting
```
c-mart: - I think you mean "play"
- The role that you call actually ensures the volume is formatted with a filesystem, "test" implies it doesn't make changes

So, I would call this "Play to ensure filesystem is present before mounting".

ansible/playbooks/utils/check_volume.yml
```diff
@@ -0,0 +1,21 @@
...
+  hosts: atmosphere
```
c-mart: You intend this to run on the Atmosphere(1) server? I would think it runs on the target instance..

ansible/playbooks/utils/mount_volume.yml
```diff
@@ -4,21 +4,18 @@
...
-      when: "{{ src is undefined }}"
+    - fail: msg="Must set VOLUME_DEVICE_TYPE. Pass a value in via `-e` as JSON {{example[0]}}"
+      when: "{{ VOLUME_DEVICE_TYPE is undefined }}"
 
-    - fail: msg="Must set fstype. Pass a value in via `-e` as JSON {{example[0]}}"
-      when: "{{ fstype is undefined }}"
 
-
-- name: Playbook to mount a volume 
+- name: Playbook to test volume before mounting
```
c-mart: What is being tested with this play? Looks like it just calls a role to mount the volume.

ansible/roles/atmo-mkfs-volume/tasks/main-alt.yml
```diff
@@ -0,0 +1,22 @@
+---
```
c-mart: Is this file dead code? If so please don't PR it.

ansible/playbooks/utils/mount_volume.yml
```diff
@@ -4,21 +4,18 @@
...
 
-    - fail: msg="Must set name. Pass a value in via `-e` as JSON {{example[0]}}"
-      when: "{{ name is undefined }}"
+    - fail: msg="Must set VOLUME_DEVICE. Pass a value in via `-e` as JSON {{example[0]}}"
+      when: "{{ VOLUME_DEVICE is undefined }}"
 
-    - fail: msg="Must set src. Pass a value in via `-e` as JSON {{example[0]}}"
-      when: "{{ src is undefined }}"
+    - fail: msg="Must set VOLUME_DEVICE_TYPE. Pass a value in via `-e` as JSON {{example[0]}}"
+      when: "{{ VOLUME_DEVICE_TYPE is undefined }}"
```
c-mart: You don't test for `VOLUME_MOUNT_LOCATION`, should we if we're testing for the other variables?

ansible/playbooks/utils/mount_volume.yml
```diff
@@ -4,21 +4,18 @@
   hosts: all
   vars:
     example:
-      - "-e \"{'name':'/vol1', 'src':'/dev/vdc', 'fstype': 'ext3'}\""
+      - "-e \"{'VOLUME_DEVICE':'/vol1', 'VOLUME_MKFS_DEVICE':True, 'VOLUME_DEVICE_TYPE': 'ext4'}\""
```
c-mart: I don't see where this playbook uses `VOLUME_MKFS_DEVICE`, but it does use `VOLUME_MOUNT_LOCATION` which isn't in your example.

ansible/roles/atmo-mount-volume/tasks/main-alt.yml
```diff
@@ -0,0 +1,22 @@
+---
```
c-mart: This file also appears to be dead code, to me.

ansible/playbooks/utils/check_volume.yml
steve-gregory: All playbooks use `hosts: atmosphere` AFAIK. 

However it's clear there is ambiguity in the naming  -- perhaps another PR to rename `atmosphere` hosts to some variant of `atmosphere-vms` to avoid the ambiguity.

ansible/playbooks/utils/check_volume.yml
steve-gregory: 👍 good suggestions here.

w.r.t. the question about `example`, presumably you could have more than one example, but I carried this along from other places in the code and figured it was some form of standard.

If there is no such standard, perhaps we should write up a ticket to remove all such instances from the codebase?

ansible/playbooks/utils/check_volume.yml
```diff
@@ -0,0 +1,21 @@
+---
+
+- name: Checking for necessary --extra_vars
+  hosts: all
+  vars:
+    example:
+      - "-e \"{'VOLUME_DEVICE':'/dev/sdb', 'VOLUME_MKFS_DEVICE':True, 'VOLUME_DEVICE_TYPE': 'ext4'}\""
```
c-mart: What do we use `VOLUME_MKFS_DEVICE` for?

steve-gregory: Ah, i pulled this variable because we won't force mkfs. Sorry, missed that!

c-mart:
Also, if we're renaming the `src` variable to `VOLUME_DEVICE` in the atmo-mount-volume role and the mount_volume.yml playbook, then should we make corresponding changes in the unmount-volume role and the unmount_volume.yml playbook?

steve-gregory:
We should, but not in this PR because unwrapping the "other" things that we do it umount like calling `lsof` on "busy" devices and bubbling up those results to the user as part of the error message, were too much to do in a single day 👍

ansible/playbooks/utils/check_volume.yml
c-mart: I don't know, the "best practice" I've seen is documenting required variables are defined at the role level, not the playbook level. But we seem to be doing things differently.

PR #80, opened 2017-04-19
Ubuntu 16.04 support

c-mart:
Issues fixed:
- Copied Previous LDAP fixes form [this branch](https://github.com/cyverse/atmosphere-ansible/tree/ubuntu1604fixes)
- had to port `hostname` DHCP exit script from `bash` to `dash` (more detail [here](https://pods.iplantcollaborative.org/jira/browse/ATMO-1555?focusedCommentId=58558&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-58630)). This fixes error where instance took a long time to reboot and did not have networking after reboot.

As part of this development/testing effort, I had to create a GUI base image for Ubuntu 16.04. We want to provide a web browser in GUI instances, but I think it's a bad idea to bake a browser into a Glance image, because it will quickly go out of date and be vulnerable to security issues, so I created a new role in atmosphere-ansible to deploy a web browser to new instances. Also found some unused Google Chrome repositories in atmosphere-ansible and removed them.

To do:
- [x] Test Ubuntu 16.04 with Jetstream-specific Ansible roles (local user account, DHCP client?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ Enable these optional configurations by setting the corresponding variable to `t
 | SETUP_IRODS_ICOMMANDS    | iRODS iCommands and iRODS FUSE client                     |
 | SETUP_REALVNC_SERVER     | RealVNC server for Atmosphere Web Desktop feature         |
 | SETUP_GLOBUS_CONNECT     | [Globus Connect](https://www.globus.org/globus-connect)   |
+| SETUP_GUI_BROWSER        | Web browser on instances with a GUI                       |
 
 ## Utility Playbooks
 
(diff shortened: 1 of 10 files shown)


c-mart:
Tested Ubuntu 16.04 image on Jetstream Test Cloud with Jetstream-specific Ansible roles. Completes to successful deployment. The one Jetstream-specific role that runs there (`atmo-local-user-account`) works as expected.

I suggest we deploy this with X-release so that we (and Jetstream) can publish Ubuntu 16.04 images. With the amount of testing I've done, it's quite possible something is broken that I'm not aware of yet, but I think the best way to proceed here is to get the images out there and fix bugs as they're discovered.

What do you think @steve-gregory?

steve-gregory:
👍 :shipit: 👍

PR #81, opened 2017-04-26
Remove idrop

calvinmclean:
This PR adds a single task to `atmo-cleanup` that deletes all files/directories that deal with iDrop.
These files are in:
- `/opt`
- `/etc/skel`
- `/home/username/Desktop`

I also removed the `atmo-idrop` role. It was already removed from playbooks.

Edit: Edwin wants this deployed to production once it is approved and merged.


c-mart:
Looks good, with an incidental observation:

It looks like the atmo-idrop role contained the `iplant_backup.sh` script which we're [currently advertising to users](https://wiki.cyverse.org/wiki/display/atmman/Backing+Up+and+Restoring+Your+Data+to+the+Data+Store). We [planned to recommend that script](https://wiki.cyverse.org/wiki/display/csmgmt/Atmosphere%280%29+Marana+Cloud+Migration+Plan) to users who need to migrate data from the old iPlant Tucson Cloud to the new Marana cloud. If the role wasn't being called then this PR doesn't break anything, it merely cleans up code that wasn't running.

@edwins we need to revisit how we're supporting data transfer between instances on separate providers -- that's a conversation for a different thread.

Back to your regularly scheduled PR.

calvinmclean:
@c-mart I could do a quick edit that adds the `iplant_backup.sh` file and a task to copy it over in `atmo-common` or one of the other roles

c-mart:
@calvinmclean let's not let the `iplant_backup.sh` matter delay merging this PR -- I vote we handle it separately. Merging now.

PR #83, opened 2017-05-12
Ezaliases

edwins:
This pull request will add two files into the /etc/profile.d directory of every instance. One file the cyverse10-utils.sh will create an alias that will setup the CYVERSE_PUBLIC_IP environment variable and allow the user to see the current public-facing ip address. The second file cyverse20.ezj-setup.sh will conveniently allow the user to install anaconda and run jupyter-notebook through aliases. To see the options, the user would simply type 'ez' on the command line.


steve-gregory:
🚅 :shipit:

PR #86, opened 2017-05-30
Fix kanki-irods role to work on Ubuntu 16.04

calvinmclean:
Ubuntu 16 images require the `g++` package to properly compile irodsclient.


c-mart:
@calvinmclean should we make `g++` a package that is only installed for Ubuntu 16 targets then, and not all Ubuntu targets?

calvinmclean:
@c-mart `g++` is still required to build the irodsclient on Ubuntu 14. I did not notice this before because `g++` was already installed on Ubuntu 14 images by the time they reach this role. So if a Ubuntu 14 image doesn't have `g++` for whatever reason, it will need it.

PR #87, opened 2017-06-07
Variable to control desktop background changes

calvinmclean:
From a conversation with @c-mart, Jetstream experiences failures with some instances because of the desktop background changes. The addition of the variable will allow Jetstream to avoid this. Also, ignore errors when these tasks fail bc it is not a fatal error.

### Todo
- [x] Add variable to other repos

### DIFF
--- a/ansible/roles/atmo-realvncserver/tasks/main.yml
+++ b/ansible/roles/atmo-realvncserver/tasks/main.yml
@@ -89,6 +89,7 @@
       command: >
         gconftool-2 -s -t string /desktop/gnome/background/picture_filename "/usr/share/backgrounds/cloud_background.jpg"
       when: ansible_distribution == "CentOS"
+      ignore_errors: yes
 
     # Note: These tasks trick the os by replacing their defaults with our own instead of properly
     # setting the background because "Unable to autolaunch a dbus-daemon without a $DISPLAY for X11"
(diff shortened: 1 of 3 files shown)


ansible/roles/atmo-realvncserver/tasks/main.yml
```diff
@@ -98,15 +99,17 @@
...
       when: ansible_distribution == "Ubuntu" and ansible_distribution_major_version < "16"
+      ignore_errors: yes
 
     - name: Change desktop background on Ubuntu 16+
       copy:
         src: /usr/share/backgrounds/cloud_background.jpg
         dest: /usr/share/backgrounds/xfce/xfce-teal.jpg
         remote_src: True
       when: ansible_distribution == "Ubuntu" and ansible_distribution_major_version >= "16"
+      ignore_errors: yes
```
steve-gregory: In other parts of the codebase we use `failed_when`.. If you are curious (like I was) what the differences were between the two of them and when they were introduced into ansible: http://docs.ansible.com/ansible/playbooks_error_handling.html#controlling-what-defines-failure

c-mart: My understanding: `failed_when: false` hides any failure, so the task returns OK/changed rather than failed in all cases. `ignore_errors: yes` will show the correct "failed" return status when a task fails, but it will not halt execution of subsequent tasks in the role/playbook.

Here, we may want to see when the task fails, but we still want the instance deployment to proceed to completion.

c-mart:
My suggestions: https://github.com/calvinmclean/atmosphere-ansible/pull/3

c-mart:
Tested on atmobeta with [my PR](https://github.com/calvinmclean/atmosphere-ansible/pull/3) to Calvin's PR. Confirmed that it works in both the case of using and not using the optional functionality. In the former case, the desktop background is set, in the latter case it is not,

PR #88, opened 2017-06-08
Problem: deploy_ready fails on invalid host key

steve-gregory:
Solution: Check the stderr results for each user tested in
'check_networking'. If all show 'host key verification failed' then
inform the user there is a misconfiguration and provide a solution to
fix the configuration.

### DIFF
--- a/ansible/roles/check_networking/tasks/main.yml
+++ b/ansible/roles/check_networking/tasks/main.yml
@@ -56,9 +56,9 @@
 - name: fail when ssh errors occur related to Host key verification failed.
   fail:
     msg: "Configuration Error: Host key verification failed for IP: {{vm_ip}} . Does your ssh/config have 'StrictHostKeyChecking no' for the IP/IP Range/*"
-  when: (default_remote_user.stderr is defined and default_remote_user.stderr == "Host key verification failed.") and
-        (centos_remote_user.stderr is defined and centos_remote_user.stderr == "Host key verification failed.") and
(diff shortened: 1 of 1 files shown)


ansible/roles/check_networking/tasks/main.yml
```diff
@@ -53,6 +53,13 @@
 # - debug: msg="ubuntu output is {{ ubuntu_remote_user.stdout }} and err is {{ ubuntu_remote_user.stderr }}"
 #   when: default_remote_user|failed and centos_remote_user|failed
 
+- name: fail when ssh errors occur related to Host key verification failed.
+  fail:
+    msg: "Configuration Error: Host key verification failed for IP: {{vm_ip}} . Does your ssh/config have 'StrictHostKeyChecking no' for the IP/IP Range/*"
+  when: (default_remote_user.stderr is defined and default_remote_user.stderr == "Host key verification failed.") and
+        (centos_remote_user.stderr is defined and centos_remote_user.stderr == "Host key verification failed.") and
+        (ubuntu_remote_user.stderr is defined and ubuntu_remote_user.stderr == "Host key verification failed.")
```
cdosborn: Should we test containment rather than `==`?

steve-gregory: 🔎 Yes we should! It was required to move to `in` rather than `==` to catch things like `Host Key mismatch`! 👍

cdosborn:
👍

PR #92, opened 2017-07-13
Change vncserver executable

calvinmclean:
Some images have `vncserver` software already installed. This caused deploy_error because the tasks intended to work with RealVNC do not work with other VNC servers.

This PR changes the `vncserver` executable from `/usr/bin/vncserver` to `/usr/bin/vncserver-virtual` so RealVNC is used even if `/usr/bin/vncserver` is linked to a different VNC server.

Tested on:
- [x] [RNAseq Alignment and Counting](https://atmo.cyverse.org/application/images/1407) (the image that brought this to our attention)
- [x] CentOS 6.8
- [x] Ubuntu 14.04

### DIFF
--- a/ansible/roles/atmo-vnc/defaults/main.yml
+++ b/ansible/roles/atmo-vnc/defaults/main.yml
@@ -6,7 +6,7 @@ background_img: 'files/atmosphere_wallpaper_v1.jpg'
 
 SET_DESKTOP_BACKGROUND: false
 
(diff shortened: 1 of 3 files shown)


calvinmclean:
After adding a task to remove `tightvncserver`, it would be successfully removed but the install of RealVNC would not create the link to `/usr/bin/vncserver` so I added a task to create this link. This has the added benefit of fixing this problem on instances that may have vncservers other than tightvnc

PR #93, opened 2017-07-17
Modify gateone-gen-sshkeys to also create Guacamole keys

calvinmclean:
This PR modifies the gateone playbook to run the play twice -- once on GateOne and once on Guacamole. The role was modified to allow for additional variables.

**The design of this new feature might seem strange, so please let me know if this is a crazy way of doing things**

### Checklist before merge:
- [x] Commit to secret repos
     - [ ] Merge changes to secret repos
- [x] Test!!!
- [ ] Approved by other contributor
- [ ] Have conversation about security on Guacamole server since it will be holding and protecting private keys


ansible/roles/gateone-gen-sshkey/tasks/main.yml
```diff
@@ -10,40 +10,63 @@
...
+        dest: "{{ GATEONE_LOCAL_PUBKEY_DIR | default('/tmp/') }}"
+        flat: yes
+    - name: Ensure correct ownership of public key
+      file:
+        path: "{{ KEY_LOC.path }}/{{ ATMOUSERNAME }}{{ KEY_LOC.prefix }}_default.pub"
+        owner: "{{ KEY_LOC.owner }}"
+        group: "{{ KEY_LOC.group }}"
+  delegate_to: "{{ play_on }}"
 
 - name: grab local copy of user ssh private key
```
c-mart: Should this say "public key"?

ansible/roles/gateone-gen-sshkey/tasks/main.yml
```diff
@@ -10,40 +10,63 @@
...
+    - name: check the existence of a given ssh key by name
+      stat:
+        path: "{{ KEY_LOC.path }}/{{ ATMOUSERNAME }}{{ KEY_LOC.prefix }}_default"
+      register: go_key
 
-- name: generate the public key
-  command: "/usr/bin/ssh-keygen -N '' -t rsa -b 2048 -f /var/lib/gateone/users/{{ ATMOUSERNAME }}/.ssh/{{ ATMOUSERNAME }}_default -C {{ ATMOUSERNAME }}"
-  when: go_key.stat.exists == False
-  delegate_to: shell
+    - name: generate the public key
```
c-mart: Should this say "generate the keypair"? Generates both public and private key?

ansible/roles/gateone-gen-sshkey/tasks/main.yml
```diff
@@ -1,7 +1,7 @@
 ---
 
 # this role expects the following variables:
```
c-mart: Let's get rid of all these comments and document the variables in readme.md.

c-mart:
Also @calvinmclean, yes, let's "Have conversation about security on Guacamole server since it will be holding and protecting private keys"!

calvinmclean:
@c-mart latest commits fix the issues you mentioned:
- [x] Change variables from dictionary style to regular
- [x] Move old comments into new README.md
- [x] Rename task and playbook
- [x] Fixed misleading task names

c-mart:
Grep your latest commit for `gateone`. You missed a few spots, some of which will cause broken-ness :)

ansible/roles/sshkey-host-access/tasks/main.yml
```diff
@@ -0,0 +1,73 @@
+---
+
+- name: ensure the existence of the user directory
+  file:
+    path: "{{ KEY_LOC_PATH }}"
+    owner: "{{ KEY_LOC_OWNER }}"
+    group: "{{ KEY_LOC_GROUP }}"
+    mode: 0700
+    state: directory
+  delegate_to: "{{ play_on }}"
```
cdosborn: Can you explain this line, shouldn't `play_on` be a host?

ansible/roles/sshkey-host-access/tasks/main.yml
```diff
@@ -0,0 +1,73 @@
...
+      register: go_key
+
+    - name: generate the keypair
+      command: >
+        /usr/bin/ssh-keygen -N '' -t rsa -b 2048 -f {{ KEY_LOC_PATH }}/{{ ATMOUSERNAME }}{{ KEY_LOC_PREFIX }}_default -C {{ ATMOUSERNAME }}
+      when: go_key.stat.exists == False
+
+    - name: set the permissions on the private key file
+      file:
+        path: "{{ KEY_LOC_PATH }}/{{ ATMOUSERNAME }}{{ KEY_LOC_PREFIX }}_default"
```
cdosborn: This value `"{{ KEY_LOC_PATH }}/{{ ATMOUSERNAME }}{{ KEY_LOC_PREFIX }}_default"` is replicated many times, which is hard to read and prone to typos. Can value be moved to `vars/defaults.yml`?

ansible/roles/sshkey-host-access/tasks/main.yml
```diff
@@ -0,0 +1,73 @@
...
+        owner: "{{ KEY_LOC_OWNER }}"
+        group: "{{ KEY_LOC_GROUP }}"
+  delegate_to: "{{ play_on }}"
+
+- name: grab local copy of user ssh public key
+  set_fact:
+    pk_file: "{{ lookup('file', '/tmp/' + ATMOUSERNAME + KEY_LOC_PREFIX + '_default.pub') }}"
+  tags: debug
+
+- name: distribute the generated gate one key to the instance
```
cdosborn: Comment is not specific to 'gate one'

ansible/roles/sshkey-host-access/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+Create and move keypairs to new instances for access through Web Shells.
+
+Role Variables
+--------------
+
+| Variable                | Required | Default | Choices                   | Comments                                 |
+|-------------------------|----------|---------|---------------------------|------------------------------------------|
+| ATMOUSERNAME            | yes      |         |                           | Username of user getting the keys        |
+| GATEONE_LOCAL_PUBKEY_DIR| no       | `/tmp/`    |                           | the directory to transfer the public key locally |
```
cdosborn: `GATEONE_LOCAL_PUBKEY...` needs to be represented generically as PUBKEY_DIR

ansible/roles/sshkey-host-access/tasks/main.yml
calvinmclean: `play_on` is a host. The vars are defined in the playbook (`41_shell_access.yml`), which runs the role twice: once for gateone and once for guacamole. On the first play, `play_on` is `shell` which is the host for gateone (defined in hosts file). In the second play, it is `guac_server` which will be added to the hosts file in our variables repos.

ansible/roles/sshkey-host-access/tasks/main.yml
calvinmclean: Good idea, I will do that

ansible/roles/sshkey-host-access/tasks/main.yml
cdosborn: Okay, just a matter of preference here, I would rename play_on to target_host. shell doesn't not make me think of a host..

ansible/roles/sshkey-host-access/tasks/main.yml
```diff
@@ -0,0 +1,73 @@
...
+      file:
+        path: "{{ FULL_PUBKEY_PATH }}"
+        owner: "{{ KEY_LOC_OWNER }}"
+        group: "{{ KEY_LOC_GROUP }}"
+  delegate_to: "{{ play_on }}"
+
+- name: grab local copy of user ssh public key
+  set_fact:
+    pk_file: "{{ lookup('file', '/tmp/' + ATMOUSERNAME + KEY_LOC_PREFIX + '_default.pub') }}"
+  tags: debug
```
cdosborn: set_fact should be avoided if possible.. 
Can you explain why a lookup is being performed here? Would prefer to just stat the path, get the stdout to see if the file exists. 

I think this task can be removed, and the lookup  be moved to the  `key` attribute of `authorized_key` below. Can you explain what should happen if the files odesn't exist.

ansible/roles/sshkey-host-access/README.md
calvinmclean: I left that in because I am not sure what `GATEONE_LOCAL_PUBKEY_DIR` is for. It is not defined in defaults and I do not see it in variable repos.

After looking into it a little more:
The next task related to this one is named `grab local copy of user ssh public key` and defines a var by finding the key in `/tmp` (which is hard-coded). Since `GATEONE_LOCAL_PUBKEY_DIR` is not defined and defaults to `/tmp/`. I think I should:
1. rename the var like you mention
2. add the variable to the task mentioned above
3. put it in `defaults/main.yml` to avoid confusion in the future

ansible/roles/sshkey-host-access/tasks/main.yml
calvinmclean: Just saw this comment after responding to another one above. I will do what I listed above to clarify and rename the `GATEONE_LOCAL_PUBKEY_DIR` variable and also move the lookup like you mention

calvinmclean: If the file does not exist, it will not be distributed to authorized_keys but there will not be a role failure

calvinmclean:
@cdosborn I fixed the things you mentioned. However, after learning a bit more about the `lookup`, I am not sure what happens if the file does not exist. I will try and find out.

calvinmclean:
```
fatal: [atmo01]: FAILED! => {"failed": true, "msg": "An unhandled exception occurred while running the lookup plugin 'file'. Error was a <class 'ansible.errors.AnsibleError'>, original message: could not locate file in lookup: /home/calvinmclean/fake_file.txt"}
```

This is what happens if the file does not exist. @c-mart @cdosborn what do you

calvinmclean:
Accidentally closed bc Linux on Macbook does strange things

ansible/roles/sshkey-host-access/tasks/main.yml
```diff
@@ -0,0 +1,67 @@
...
+    - name: Ensure correct ownership of public key
+      file:
+        path: "{{ FULL_PUBKEY_PATH }}"
+        owner: "{{ KEY_LOC_OWNER }}"
+        group: "{{ KEY_LOC_GROUP }}"
+  delegate_to: "{{ target_host }}"
+
+- name: distribute the generated pub key to the instance
+  authorized_key:
+    key: "{{ lookup('file', LOCAL_PUBKEY_DIR + ATMOUSERNAME + KEY_LOC_PREFIX + '_default.pub') }}"
```
cdosborn: Does the following work: `lookup('file', FULL_PUBKEY_PATH)`. You do not need to handle failure with `lookup` since you're generating that key, and if it fails to generate, then the error will be caught there.

calvinmclean: That wouldn't work because `FULL_PUBKEY_PATH` is the path on the remote server, and this task is looking locally.

cdosborn: Okay, sounds good.

calvinmclean:
@c-mart @cdosborn any more changes I should make? Chris, let me know when you would be able to discuss the security concerns.

ansible/roles/sshkey-host-access/tasks/main.yml
```diff
@@ -0,0 +1,67 @@
...
+    state: present
+    user: "{{ item }}"
+  with_items:
+    - "{{ ATMOUSERNAME }}"
+  tags:
+    - ssh-keys
+    - debug
+
+- name: remove user key from controller node
+  local_action: file path=/tmp/{{ ATMOUSERNAME }}_default.pub state=absent
```
cdosborn: This path needs to be a variable. Isn't `tmp` configurable above, so that a change in conf, would mean this key isn't deleted?

cdosborn: LOCAL_FULL_PUBKEY_PATH = LOCAL_PUBKEY_DIR + ATMOUSERNAME + KEY_LOC_PREFIX + '_default.pub'

cdosborn: the fetch above could use this var also i think

calvinmclean: I think you're right, the key wouldn't be deleted if conf is changed to not use `/tmp`. I added the LOCAL_FULL_PUBKEY_PATH variable, but do not see a place for it in the `fetch` task.

ansible/roles/sshkey-host-access/defaults/main.yml
```diff
@@ -0,0 +1,12 @@
+---
+LOCAL_PUBKEY_DIR: "/tmp/"
```
cdosborn: By convention DIRs should not include a trailing slash

ansible/roles/sshkey-host-access/README.md
```diff
@@ -0,0 +1,34 @@
...
+| LOCAL_PUBKEY_DIR        | yes      | `/tmp/` |                           | the directory to transfer the public key locally |
+| KEY_LOC_PATH            | yes      |`/var/lib/gateone/users/{{ ATMOUSERNAME }}/.ssh`|| the directory to transfer the public key |
+| KEY_LOC_OWNER           | yes      | root    | "root", "tomcat7"         | owner of new keys                        |
+| KEY_LOC_GROUP           | yes      | root    | "root", "tomcat7"         | group of new keys                        |
+| KEY_LOC_PREFIX          | yes      | ""      | "_guac", ""               | prefix to name key for specific server   |
+| target_host             | yes      | shell   | "shell", "guac_server"    | hostname for delegated tasks             |
+| FULL_PUBKEY_PATH        | yes      |         |                           | simplifies long string of other variables|
+| FULL_PRIVKEY_PATH       | yes      |         |                           | simplifies long string of other variables|
+| LOCAL_FULL_PUBKEY_PATH  | yes      |         |                           | simplifies long string of other variables|
```
cdosborn: I submitted some changes to illustrate what i mean. 
https://github.com/calvinmclean/atmosphere-ansible/pull/5

ansible/roles/sshkey-host-access/README.md
```diff
@@ -0,0 +1,34 @@
+sshkey-host-access
+=========
+
+Create and move keypairs to new instances for access through Web Shells.
```
cdosborn: I included an example in the readme -> calvinmclean#5

ansible/roles/sshkey-host-access/defaults/main.yml
```diff
@@ -0,0 +1,12 @@
...
+
+KEY_LOC_PATH: "/var/lib/gateone/users/{{ ATMOUSERNAME }}/.ssh"
```
cdosborn: Use `*_DIR` name to refer to a directory `*_PATH` to refer to a path.. unspoken convention that is inconsistently applied :)

calvinmclean:
I deployed an instance earlier and it worked

calvinmclean:
Is this going to be merged for v27?

PR #97, opened 2017-08-15
ATMO-1992: Change ownership of mounted volume to the user

calvinmclean:
A user mentioned that their mounted volume was owned by root and they could not access it. Luckily, the user knew how to use `chown` to change the ownership, but noted it as an inconvenience.
I think this would be likely to confuse users that do not have much Linux experience.

Also removed the unused `mount-volume` role and changed the `name` attribute in the `mount` task to `path` because it was added in Ansible 2.3 and is more clear than `name`.

- [ ] Requires PR to secrets repo to be merged


ansible/roles/atmo-mount-volume/tasks/main.yml
```diff
@@ -16,9 +15,17 @@
...
   debug:
     var: mount_result
+
+- name: Change directory ownership
+  file:
+    state: directory
+    path: "{{ VOLUME_MOUNT_LOCATION }}"
+    recurse: yes
+    owner: "{{ ATMOUSERNAME }}"
+    group: "iplant-everyone"
```
steve-gregory: Lets not hard-code the group here, we will want a different value for other deployers of `atmosphere-ansible`

calvinmclean: Good point. Is there a way to do this without having to add a new variable to all the secret repos?

steve-gregory:
THanks @calvinmclean !

PR #98, opened 2017-08-23
WIP: Allow hotplug for network interfaces on Ubuntu instances

c-mart:
To do: figure out why CentOS instances do not have networking on resume from suspend


c-mart:
This doesn't work, closing, will try a different approach tomorrow.

Issue #99, opened 2017-08-24
Limit `chown` & `chmod` to a single filesystem


julianpistorius:
Otherwise network filesystems mounted in home directory will cause instance resume to fail catastrophically. Not even mentioning breaking file permissions on remote network file systems.

calvinmclean:
This is fixed by PR #100 and can be closed.

PR #100, opened 2017-08-24
Skip filesystems mounted in home directory when using "chown" or "chmod"

calvinmclean:
Fixes Issue #99 by using Ansible facts to loop through mounts and check if any are in the user's home directory. If a volume is mounted in a home directory, a fact is set so it can be unmounted and remounted after `chown`-ing home directory.

### TODO:
- [x] Test test test

### DIFF
--- a/ansible/roles/add-user/meta/.galaxy_install_info
+++ b/ansible/roles/add-user/meta/.galaxy_install_info
@@ -1,1 +1,1 @@
-{install_date: 'Mon Jul 24 21:37:45 2017', version: master}
+{install_date: 'Tue Aug 29 18:21:05 2017', version: master}
--- a/ansible/roles/add-user/tasks/main.yml
+++ b/ansible/roles/add-user/tasks/main.yml
@@ -140,14 +140,12 @@
   shell: cp -R /etc/skel /home/"{{ ATMOUSERNAME }}"
   when: not atmouser_home.stat.exists
 
-# This chown will cause errors if /home/ATMOUSERNAME/.gvfs is mounted.  We should add code to manually unmount it if
-# it exists, then chown that user's home. -mgd
-- name: set ownership for user's home
-  file:
-    path: "/home/{{ ATMOUSERNAME }}"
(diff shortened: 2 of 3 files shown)


julianpistorius:
@calvinmclean I suspect you're going to run into the problem where a process (or processes) lock a volume from being unmounted, and then resume will fail.

I couldn't find documentation for limiting `chmod` & `chown` to a single file system, but maybe if you combine it with the `find` command and limiting it using `-mount`:

```
-mount Don't descend directories on other filesystems.  An alternate name for -xdev, for compatibility with some other versions of find.
```

calvinmclean:
Do we want to avoid killing those processes? `fuser -c -k /mount-point` will kill anything accessing `/mount-point`

julianpistorius:
We should _definitely_ avoid killing processes.

Imagine a long-running process which can't be restarted, and which has to start from the beginning. Or something which could create corrupted data if interrupted forcibly.

julianpistorius:
A command like this should do it:

```bash
find /home/johndoe -mount -exec chown johndoe {} +
```

julianpistorius:
Though you should probably use `xargs` - or that command above will fail on home directories with many files:

```bash
find /home/johndoe -mount | xargs -n 3 chown johndoe
```

Note: Remove the `-n 3` part. It's only to demonstrate the batch sizes.

julianpistorius:
I was hoping that the `follow` attribute  might be all you need, but I suspect that only applies to (symbolic?) links, and not file systems. But it is worth trying:

http://docs.ansible.com/ansible/latest/file_module.html

parameter | required | default | choices | comments
---------- | --------- | ------- | ------- | -----------
follow(added in 1.8) | no | no | yesno | This flag indicates that filesystem links, if they exist, should be followed.

calvinmclean:
Tested and confirmed that with `follow: no`, filesystems are still followed. I'll try out the xargs option next.

julianpistorius:
Eeeexcellent. Good to hear.

steve-gregory:
Thanks @calvinmclean !

PR #103, opened 2017-09-07
Playbooks to replace libcloud deployments

steve-gregory:
These playbooks have been created to replace the legacy libcloud `ScriptDeployment` in Atmosphere.


steve-gregory:
Self-:shipit:

PR #104, opened 2017-09-08
Improve the boot scripts experience

steve-gregory:
- Create a directory for 'post-scripts' (Baked into the image)
  and for 'instance-scripts' (added to the instance)
- Write a shell command that will be the 'script runner':
  - Creates dynamic log location for each script, including datetime for
  multiple runs.
  - 'Fire and forget' scripts to avoid 'hanging' scripts

Built in conjunction with https://github.com/cyverse/atmosphere/pull/506

### DIFF
--- a/ansible/roles/atmo-user-boot-scripts/tasks/main.yml
+++ b/ansible/roles/atmo-user-boot-scripts/tasks/main.yml
@@ -1,38 +1,43 @@
 - name: bootscript_1 - make directories for post scripts and logs
   file: path={{ item }} state=directory
-  with_items: '{{ POST_MAKE_DIR }}'
+  with_items: '{{ POST_MAKE_DIRS }}'
 
-- name: bootscript_2 - touch log files
-  file: path={{ item }} state=touch
-  with_items: '{{ POST_SCRIPT_LOGS }}'
-  # The following two lines work around https://github.com/ansible/ansible-modules-core/issues/170
-  register: touch_log
-  changed_when: touch_log.diff.before.state != "file"
-
-- name: bootscript_3 - run user's scripts in script directory
-  shell: >
-    for SCRIPT in {{ POST_SCRIPT_DIR_RUN }};
-    do
-        if [ -f $SCRIPT -a -x $SCRIPT ]
-        then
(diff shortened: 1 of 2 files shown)


ansible/roles/atmo-user-boot-scripts/tasks/main.yml
```diff
@@ -1,38 +1,43 @@
...
+        SCRIPT_STDOUT=$LOG_DIR/$BASENAME.$DATE.stdout
+        SCRIPT_STDERR=$LOG_DIR/$BASENAME.$DATE.stderr
+        ATMO_USER={{ATMOUSERNAME}} nohup $SCRIPT >$SCRIPT_STDOUT 2>$SCRIPT_STDERR </dev/null &
+      fi
+    done
+    exit 0
+  args:
+    executable: /bin/bash
+  register: shell_result
+  with_items: '{{ RUN_SCRIPTS }}'
```
steve-gregory: Noted by @julianpistorius -- In the past we used to tell users that their boot scripts had failed.

In this new method (fire and forget) we will not have that ability..

Some suggestions:
1) go with fire and forget.. Update documentation and guides appropriately
2) "fire, forget, and monitor" Track the pids? wait for the script to exit, using the return code on script-completion to denote success/failure (if you can even do that?) and then setting `user_deploy_error` accordingly based on the failure.
3) "Monitor the scripts": Remove async logic and fail out when the script returns a non-zero code. Allow the error to bubble up and surface a `user_deploy_error`

ansible/roles/atmo-user-boot-scripts/tasks/main.yml
```diff
@@ -1,38 +1,71 @@
...
+    content: "{{ item.text }}"
+    dest: "{{INSTANCE_SCRIPT_DIR}}/{{item.name}}"
+    mode: "u+rwx"
+  with_items: '{{ ASYNC_SCRIPTS }}'
+
+#Future TODO:
+# - If possible, refactor these components into ansible:
+#   - get name of  each file in script directory
+#   - if file is executable, generate 'dynamic log files' based on 'script|basename' and 'date()'
+#   - if file is executable, run ` set -m; nohup-call-to-file..` and point to dynamic log files from above.
```
steve-gregory: Drawing attention to this TODO for anyone aspiring to make such a change. Otherwise, the code and can be used as-is.

steve-gregory:
Self-:shipit:

PR #106, opened 2017-09-12
WIP ATMO-1978: "check_web_desktop: True" when NOVNC is disabled

calvinmclean:
It looks like Atmosphere checks NOVNC by running a playbook and looking for failures. When `SETUP_NOVNC = false`, the playbook skips the role, so there are no failures. This PR adds a task to be executed before the roles, and will fail if `SETUP_NOVNC = FALSE`.

#### TODO:
- [ ] Needs to be tested
This has pretty specific testing circumstances so I haven't been able to test yet.

### DIFF
--- a/ansible/playbooks/utils/atmo_check_novnc.yml
+++ b/ansible/playbooks/utils/atmo_check_novnc.yml
@@ -1,5 +1,10 @@
 ---
 - name: Playbook to fail when VNC is not running
   hosts: atmosphere
+  pre_tasks:
+  - name: Fail when SETUP_NOVNC is false because the role will skip instead of fail
+    fail:
+      msg: "SETUP_NOVNC is false or undefined"
+    when: SETUP_NOVNC is not defined or SETUP_NOVNC == false
   roles:
     - { role: atmo-check-novnc, when: SETUP_NOVNC is defined and SETUP_NOVNC == true }


cdosborn:
> When SETUP_NOVNC = false, the playbook skips the role, so there are no failures.
Why is that an issue?

calvinmclean:
Atmosphere looks for failures while running the playbook, so if there are no failures, Atmosphere will set `check_web_desktop` to `True`. My understanding of this is that the "Web Desktop" action button will then be added to instances even when `SETUP_NOVNC == false`.

cdosborn:
Okay. it would have been helpful to me to include a problem statement like: When SETUP_NOVNC = false, NoVNC support appears in the browser.

This fix works but is applied to the wrong component. The test shouldn't know whether it should be run or skipped, the test has one job: see if vnc's going to work out.

The desired functionality is that an atmo deployment can configure whether or not instances have web desktop (i.e no_vnc) enabled. Just like how deployments can configure if guacamole is supported. After looking at atmosphere, I found that this support doesn't exist for web desktop. I'll update the jira with the info that I found. The gist is that atmosphere should support a "WEB_DESKTOP_ENABLED" flag, and that flag should determine if we ever even run the "check_no_vnc" playbook

PR #107, opened 2017-09-12
WIP ATMO-1979: Atmosphere-Ansible adds/changes VNC on VMs with X2Go

calvinmclean:
Some images are designed to be used with X2Go for visualization, but ~~the `atmo-vnc` role wipes out that functionality~~ it prevents Atmosphere Web Desktops from working properly. This PR adds a pre_task to the playbook that executes the `atmo-vnc` role to make the role only run when `/etc/x2go` does **not** exist.

#### TODO:
- [ ] Needs some more testing. Works as a stand-alone playbook, but I have to test it in the atmosphere-ansible workflow.

### DIFF
--- a/ansible/playbooks/instance_deploy/30_post_user_install.yml
+++ b/ansible/playbooks/instance_deploy/30_post_user_install.yml
@@ -2,11 +2,16 @@
 # Requires ATMOUSERNAME and VNCLICENSE
 - name: Playbook lists roles that should be executed after ATMOUSERNAME has been created
   hosts: atmosphere
+  pre_tasks:
+    - name: Determine if X2GO is installed
+      stat:
+        path: "/etc/x2go"
+      register: x2go
   roles:
     - { role: atmo-fail2ban, when: true }
     - { role: irods-icommands, when: SETUP_IRODS_ICOMMANDS is defined and SETUP_IRODS_ICOMMANDS == true }
     - { role: atmo-kanki-irodsclient, when: SETUP_IRODS_ICOMMANDS is defined and SETUP_IRODS_ICOMMANDS == true }
     - { role: install-browser, when: SETUP_GUI_BROWSER is defined and SETUP_GUI_BROWSER == true }
-    - { role: atmo-vnc, when: SETUP_REALVNC_SERVER is defined and SETUP_REALVNC_SERVER == true }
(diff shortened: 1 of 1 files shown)


cdosborn:
Never heard of x2go. Why does installing atmovnc prevent that from working? My concerns are that this is adding unnecessary edge case handling, and disabling part of the expected experience (having vnc support through the ui). A nice fix would be to change the vnc to not stomp out x2go if thats possible. It's a win if our codebase doesn't know about x2go. If you could include some more info like how popular x2go is that would help frame it. Its possible that we should definitely support x2go.

calvinmclean:
I never heard of it either. Just trying to take care of some small bugs in JIRA. X2Go still functions as expected after the instance is deployed, so we aren't preventing people from using X2Go, but instances with X2Go prevent our Web Desktops from working properly. My original description is misleading, edited it now.

I agree that this is a strange edge case (I think it only applies to BioLinux images, according to JIRA ticket). However, I think cases like this should be handled in _some_ way (not necessarily this way) because it makes it look like Atmosphere is broken to the average user that may not know X2Go is on the instance. And users that intend to use X2Go won't be looking for Web Desktop anyways.

I'll try and see if there is a nice way to run VNC and X2Go side-by-side.

c-mart:
I've used X2Go quite a lot, and in a previous job we evaluated it for use by a bunch of people to access their virtualized workstation. When I did that research a couple years ago, I found that X2Go was the overall best option (in both performance and freedom) for remotely accessing a Linux desktop environment. (X2Go uses [NX technology](https://en.wikipedia.org/wiki/NX_technology#Brief_history_of_NX) after original developer NoMachine took the code in a proprietary and backwards-incompatible direction.)

I agree with @cdosborn that we should avoid this kind of edge case handling when we can. Also, presence of `/etc/x2go` is not a reliable indication that the X2Go service is actually enabled or running. Perhaps VNC and X2Go servers just have to use different displays or listen on different ports, and they can co-exist.

calvinmclean:
I don't think running them on different displays will work because Guacamole and NoVNC run on different displays, but they are both disrupted by X2Go. Maybe we could add a task to disable all non-VNC display services like X2Go and XSPICE. This would be more generic and solve the same problem. Users would then have the option to re-enable those if they need them.

PR #102, opened 2017-09-06
Modifications to /etc/hosts at Andy E's suggestion

c-mart:
Follow-up on #101, setting `/etc/hosts` to include the external IP address. Test results after deploying from atmobeta follow.

On an "Ubuntu 16.04 Non-GUI Base":
```
root@vm142-35:~# hostname
vm142-35.cyverse.org
root@vm142-35:~# hostname -s
vm142-35
root@vm142-35:~# uname -n
vm142-35.cyverse.org
root@vm142-35:~# cat /etc/hostname
vm142-35.cyverse.org
root@vm142-35:~# cat /etc/hosts
127.0.0.1 localhost

# The following lines are desirable for IPv6 capable hosts
::1 ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
ff02::3 ip6-allhosts
128.196.142.35 vm142-35.cyverse.org vm142-35
```

On an "Ubuntu 14.04.3 NoGUI Base":
```
root@vm142-72:~# hostname
vm142-72.cyverse.org
root@vm142-72:~# hostname -s
vm142-72
root@vm142-72:~# uname -n
vm142-72.cyverse.org
root@vm142-72:~# cat /etc/hostname
vm142-72.cyverse.org
root@vm142-72:~# cat /etc/hosts
127.0.0.1 localhost

# The following lines are desirable for IPv6 capable hosts
::1 ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
ff02::3 ip6-allhosts
128.196.142.72 vm142-72.cyverse.org vm142-72
```

On "CyVerse CentOS 6.8 GUI Base":
```
-bash-4.1# hostname
vm142-62.cyverse.org
-bash-4.1# hostname -s
vm142-62
-bash-4.1# uname -n
vm142-62.cyverse.org
-bash-4.1# cat /etc/hostname
vm142-62.cyverse.org
-bash-4.1# cat /etc/hosts
127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4
::1         localhost localhost.localdomain localhost6 localhost6.localdomain6
128.196.142.62 vm142-62.cyverse.org vm142-62
```

[...]

PR #108, opened 2017-09-15
Include short hostame in /etc/hosts

afgane:
This complements #101 (and is needed when Slurm is being installed on the system, for example)


c-mart:
@afgane I believe this PR duplicates a similar change in #102. Just waiting on some review there. What do you think?

afgane:
#102 will work (just confirmed/tested it) so I'll close this.

PR #109, opened 2017-09-18
Restart networking for CentOS 6 on deploy

c-mart:
Possible fix for CentOS 6 instances having broken networking after suspend+resume. Tried to test but atmobeta is not in a great state at the moment, will revisit.

### DIFF
--- a/ansible/playbooks/instance_deploy/10_setup_pkg_mgr.yml
+++ b/ansible/playbooks/instance_deploy/10_setup_pkg_mgr.yml
@@ -3,6 +3,7 @@
   hosts: atmosphere
   roles:
     - atmo-pre-setup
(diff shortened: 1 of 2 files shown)


c-mart:
Tested on atmobeta:
- Launched a `CyVerse CentOS 6.8 GUI Base`, waited for green-light active status
- Confirmed I can SSH in, set root password for potential console troubleshooting
- Suspended instance, then resumed instance
- After a minute, instance received new floating IP address. Confirmed that:
  - I can SSH to instance using new floating IP
  - Instance also successfully got a new DHCP lease with different internal/'fixed' IP address

Removing "WIP" label.

steve-gregory:
Thanks @c-mart :shipit:

PR #111, opened 2017-09-18
Don't start vncserver if its already running

cdosborn:
Problem: deploy failure, vncserver can not be started if its already running
on a display
Solution: Use a command to test if vnc is already running. Then start it if
it's not running.


cdosborn:
Im a little worried about the usage of `vnc-server -test`. It exhibited the right behavior, but the docs suggest it does something else.

Docs say:
```
    -test            Print Xvnc command-line without executing.
```
You would think that would never fail, but it fails if  vncserver is running on the display already.

cdosborn:
I tested this against a vm exhibiting the issue.

PR #112, opened 2017-09-20
check-restart-network script

c-mart:
Apparently I was mistaken with #109, and more is needed to bring CentOS 6 instances back online after a suspend + resume. Let me test this one more time.


c-mart:
Tested with two consecutive instances on atmobeta. Suspended, resumed, waited a few minutes, were SSHable again. This should do the trick.

steve-gregory:
👍 Thanks @c-mart :shipit:

PR #115, opened 2017-09-27
Provide a default value for SETUP_GUACAMOLE

cdosborn:
Problem: Missing variable results in instance deploy failure
Solution: Don't require variable to be defined in the secrets repo, provide a default.

When you deploy, if `SETUP_GUACAMOLE` is missing from the secrets repo group_vars file, atmosphere-ansible fails on instance deploy, because the variable is undefined. This variable was missing from jtc group_vars. Rather than require it to be added, just provide a default.


steve-gregory:
Thanks @cdosborn !

PR #116, opened 2017-09-28
WIP - Use atmosphere variables over group_vars, match variable names

cdosborn:
Problem: Variables are duplicated in group_vars and in atmosphere and have
different names
Solution: Use atmosphere naming, requires atmo to pass new vars to supspace

This makes the SSHKEYS and SSHKEYS_TO_REMOVE no longer necessary. Requires
atmosphere to call subspace with ATMOSPHERE_KEYPAIR_FILE and
ATMOSPHERE_PRIVATE_KEYFILE, which are defined in atmosphere.settings;

~~Depends on  https://github.com/cyverse/atmosphere/pull/523~~

Depends on `<pending clank pr>`

### DIFF
--- a/ansible/roles/add-user/defaults/main.yml
+++ b/ansible/roles/add-user/defaults/main.yml
@@ -1,23 +1,9 @@
 ---
 
-SUDOERS_USERS_AND_GROUPS:
-  TO_ADD:
-    users:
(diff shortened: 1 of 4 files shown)


cdosborn:
Secrets repos no longer need to define: `SSHKEYS` and `SSHKEYS_TO_REMOVE`

PR #118, opened 2017-10-03
Changes staged but not commited on atmo

steve-gregory:


steve-gregory:
Closing -- no changes left, master includes all necessary changes.

PR #121, opened 2017-10-10
Disable serving RealVNC Java Viewer from instance

c-mart:
Fix for [JETSTREAM-129](https://iujetstream.atlassian.net/projects/JETSTREAM/issues/JETSTREAM-129?filter=allopenissues). Tested on atmobeta and confirmed the error no longer occurs with an Ubuntu 16.04 GUI XFCE Base.


steve-gregory:
Thanks @c-mart !

PR #122, opened 2017-10-10
WIP ATMO-1969: Modify conditionals in atmo-ssh-setup to avoid 'undefined' errors.

calvinmclean:
This PR addresses a failure that would occur when variables like `root_connection` are defined, but `root_connection.stdout` is not. This is the error: `Make sure your variable name does not contain invalid characters like '-': argument of type 'StrictUndefined' is not iterable`. This error is not clear, but after some online research and trying different variable combinations, I confirmed that it is caused by the undefined `root_connection.stdout`.

In order to check if `root_connection.stdout` is undefined, we should first check if `root_connection` is undefined, because if it is, then `root_connection.stdout is defined` will fail.

The JIRA ticket says that atmosphere-ansible fails after N number of deployments, but I was unable to re-create this error because of the unknown N. However, I confirmed that instances still deploy with these modified conditionals.

### DIFF
--- a/ansible/roles/atmo-ssh-setup/tasks/main.yml
+++ b/ansible/roles/atmo-ssh-setup/tasks/main.yml
@@ -17,29 +17,36 @@
       command ssh {{ SSH_OPTIONS }} -p {{ ansible_port }} centos@{{ vm_ip }} echo "Atmosphere is cool!"
     register: centos_connection
     failed_when: false
-    when: "'Atmosphere is cool!' not in root_connection.stdout"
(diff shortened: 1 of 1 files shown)


calvinmclean:
If first attempt (root_connection) fails, and second attempt (centos_connection) succeeds, the third (ubuntu_connection) will still be attempted. This does not result in errors because the `use_remote_user` is still correctly set to `centos`.

This also happens when using the master branch.

calvinmclean:
Closed in favor of #128

PR #123, opened 2017-10-20
Problem: atmosphere-ansible reliance on ssh_config can make it hard to debug...

steve-gregory:
... sometimes different SSH keys are used, etc.

Solution: Update SSH Options to avoid reliance on ssh config

**WIP: This needs to become configurable before merging, because not all paths will match this.**


steve-gregory:
Development moved to https://github.com/cyverse/atmosphere-ansible/pull/124

PR #124, opened 2017-10-20
Problem: atmosphere-ansible ssh setup tasks can be hard to debug

steve-gregory:
Solution: 
- Pass in all expected SSH args as options.
- Allow `atmosphere` to set the `SSH_IDENTITY_FILE` in extra_vars.


c-mart:
Merging with two comments: 
- We're already defining SSH_IDENTITY_FILE in a various other places, perhaps it should be set here from elsewhere
- At some point we may want to set `SSH_OPTIONS` in only one place (e.g. `group_vars`)

PR #125, opened 2017-10-25
Mount ephemeral disk to /scratch instead of /home, provide warning message

c-mart:
We are no longer mounting `/home` to ephemeral storage. If an ephemeral disk is provided, launch it to `/scratch` instead, and provide a data loss warning to the user who might put things there.

Tested in my local dev environment against a production cloud.


ansible/roles/atmo-ephemeral-mount/defaults/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+
+MOUNT_SOURCE: 'vdb'
```
steve-gregory: Can we dynamically figure this out?

ansible/roles/atmo-ephemeral-mount/tasks/main.yml
```diff
@@ -0,0 +1,21 @@
...
+  when: 'item.mount == "/home"'
+  with_items: '{{ ansible_mounts }}'
+
+- name: 'Ephemeral disk mounted'
+  mount:
+    path: '{{ MOUNT_PATH }}'
+    src: '/dev/{{ MOUNT_SOURCE }}'
+    fstype: '{{ MOUNT_TYPE }}'
+    state: 'mounted'
+  when: '(ansible_devices.{{ MOUNT_SOURCE }} is defined) and (home_mounted is not defined)'
```
steve-gregory: 👍

ansible/roles/atmo-ephemeral-mount/defaults/main.yml
c-mart: Maybe, but in my experience, an instance launched with an ephemeral disk will always have the ephemeral disk at [`/dev/vdb`](https://docs.openstack.org/glance/pike/admin/troubleshooting.html).

One more thing I forgot about: if you launch an instance using a flavor that has no ephemeral disk (e.g. on Jetstream Cloud), then attach a volume, that volume will also appear as `/dev/vdb`, so on redeploy of the same instance, the `atmo-ephemeral-mount` role will probably fail with error (because `/deb/vdb` is already mounted at `/vol1` or whatever).

Pushing another commit to fix this in a moment.

c-mart: Fixed.

ansible/roles/atmo-ephemeral-mount/tasks/main.yml
```diff
@@ -1,33 +1,44 @@
...
+- name: 'Determine if ephemeral disk is already mounted in a non-default location'
+  # If it is, we won't alter it
+  set_fact:
     disk_mounted_elsewhere: true
-  when: '(item.device == "/dev/{{ MOUNT_SOURCE }}") and (item.mount != "/mnt")'
+  when:
+    - 'ephem_stat.islnk is defined'
+    - 'ephem_stat.islnk == true'
+    - 'item.device == "{{ ephem_stat.lnk_source }}"'
+    - 'item.mount != "/mnt"'
```
steve-gregory: First time I've seen this syntax :+1:

c-mart:
Tested, I believe it's ready to merge.

steve-gregory:
Thank you @c-mart :shipit:

Issue #105, opened 2017-09-11
Readme is out of date


cdosborn:
The sections about playbooks is wrong. The playbooks have been moved into sub directories.

steve-gregory:
Fixed by #127.

PR #127, opened 2017-10-26
Update README with new playbook directories and names

calvinmclean:
This is a fix for issue #105 

Recently we changed the directory structure of our playbooks because there were too many for just the one directory. The README has been updated to correctly show the deploy playbooks and sections are added for new playbook categories.


steve-gregory:
:shipit: Thank you @calvinmclean

PR #129, opened 2017-11-09
Set default font for XTerm to TrueType

calvinmclean:
Ubuntu 16 instances with XTerm default to a small, hard to read font. The problem is exaggerated when the display resolution is increased. This PR changes the default font size and enables rendering for TrueType fonts. Here is an image with terminals before and after changing the defaults (display resolution is 1920x1080):

screen shot 2017-11-09 at 11 14 26 am

Things I noticed during testing:
- CentOS and Ubuntu 14 instances are unaffected by the changes because the default terminal emulator is not XTerm. This is fine because the default terminal is good looking and has intuitive menus for customization if needed.

### DIFF
--- a/ansible/roles/atmo-vnc/files/Xresources
+++ b/ansible/roles/atmo-vnc/files/Xresources
@@ -1,0 +1,3 @@
+xterm*faceName: Ubuntu Mono
+xterm*faceSize: 12
+xterm*renderFont: true
--- a/ansible/roles/atmo-vnc/tasks/main.yml
+++ b/ansible/roles/atmo-vnc/tasks/main.yml
@@ -65,4 +65,6 @@
 
   - include: novnc.yml
     when: SETUP_NOVNC is defined and SETUP_NOVNC == true
+
+  - include: xterm_font.yml
   when: has_gui
--- a/ansible/roles/atmo-vnc/tasks/xterm_font.yml
+++ b/ansible/roles/atmo-vnc/tasks/xterm_font.yml
@@ -1,0 +1,16 @@
+---
+
+- name: Create .Xresources file for user
+  copy:
+    src: 'files/Xresources'
+    dest: '/home/{{ ATMOUSERNAME }}/.Xresources'
+
+- name: Update X server for all displays
+  become: true
+  become_user: '{{ ATMOUSERNAME }}'
+  command: >
+    /usr/bin/xrdb -display {{ item }} -merge /home/{{ ATMOUSERNAME }}/.Xresources
+  with_items:
+    - ':1'
+    - ':2'
+    - ':5'


lenards:
Can we get a reviewer on this? I know from my support service, the Xterm font not being a TrueType is a significant problem for the readability of the terminal text.

ansible/roles/atmo-vnc/files/Xresources
```diff
@@ -0,0 +1,3 @@
+xterm*faceName: Ubuntu Mono
+xterm*faceSize: 12
+xterm*renderFont: true
```
steve-gregory: LG2M -- As long as these keynames (`xterm*...`) are valid to ansible, i say we go ahead and :shipit:

lenards: Do we have a fallback fontface for CentOS/Red Hat?

lenards: I realize that there was testing to see that this does *not* impact CentOS / Ubuntu 14:

> CentOS and Ubuntu 14 instances are unaffected by the changes because the default terminal emulator is not XTerm. This is fine because the default terminal is good looking and has intuitive menus for customization if needed.

But I would be curious if there is a way to use _facts_ to select a fontface more universally (perhaps).

c-mart:
xterm is not a super user-friendly terminal app regardless. Instead, I suggest we install `xfce4-terminal` on instances with an xfce window manager? Then the user gets a nice font picker in the preferences menu, and probably a nicer default font too. Comparison on my box (which runs xfce):

image

Also, I think putting these changes in the `atmo-vnc` role overloads it; would suggest making a `gui-terminal` role for these kinds of tweaks.

calvinmclean:
@c-mart I agree, I always install konsole on my instances because I cannot stand xterm. I will get started on making it install `xfce4-terminal` on xfce instances. And I will do it in a new role.

@lenards since we are probably going to opt for Chris' suggestion instead of changing font, we don't have to worry about the "Ubuntu Mono" font. I will look at our CentOS instances and see if they will also need a terminal upgrade.

calvinmclean:
@c-mart what would you think about changing the name of the role `install-browser` to something like `atmo-gui-tweaks`, and then adding a new task file called `terminal.yml` that is included from `main.yml`? Then we could also consider moving some other tasks from `atmo-vnc`, like changing desktop background, to task files in `atmo-gui-tweaks`. I think it is a clean solution that will take some of the load off the `atmo-vnc` task since it handles a lot already with guacamole/novnc/realvnc.

c-mart:
Sorry that I missed these replies. @calvinmclean that sounds reasonable to me.

PR #130, opened 2017-11-09
Create desktop shortcut to change the display resolution

calvinmclean:
The default display resolution for VNC sessions is 1024x768, which does not give the user much room to work with. If users browse through the wiki they would be able to find the command to change the resolution, but I assume many don't know that is an option. This PR will allow users to easily choose from all the available resolution options.


calvinmclean:
Closing this so I can redo it as a gui-tweak when #137 is merged.

PR #131, opened 2017-11-13
Problem: Currently generated config file throws deprecation warnings

steve-gregory:
### Highlight Reel:
- Note where specific settings will be deprecated (including the version of ansible)
- Fix some old defaults to match those listed in the [ansible configuration file document](http://docs.ansible.com/ansible/latest/intro_configuration.html).
- Attempt to include additional 'sections' within the `[general]` config directive (Could be re-refactored in the future)

### DIFF
--- a/ansible/ansible.cfg.j2
+++ b/ansible/ansible.cfg.j2
@@ -1,64 +1,83 @@
 # config file for ansible -- http://ansible.com/
 # ==============================================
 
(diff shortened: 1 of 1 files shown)


steve-gregory:
Seeking review -- Otherwise I'll self-:shipit: on Monday

cdosborn:
Fixed in https://github.com/cyverse/atmosphere-ansible/pull/155

PR #132, opened 2017-11-21
Problem: remove-user is not idempotent.

steve-gregory:
Solution: Make remove-user idempotent by verifying the user was removed from `/etc/passwd` and ignoring failures from `userdel --remove <username>`


cdosborn:
Can you include a link + comment to explain the workaround. Looks like a bug in ansible?

PR #133, opened 2017-12-12
Ubuntu only: fixes false positive ban of users who successfully login too many times

edwins:
This pull request essentially changes the order of of pam auth to ldap first, then local auth second


lenards:
👏 :shipit:

PR #134, opened 2018-01-12
Problem: instances fail to boot if a volume is removed outside of Atmosphere

calvinmclean:
If a volume is removed without using the [atmosphere-ansible unmount_volume playbook](https://github.com/cyverse/atmosphere-ansible/blob/master/ansible/playbooks/instance_actions/unmount_volume.yml), the instance will fail to boot after being rebooted/redeployed/resumed because a disk in `/etc/fstab` is missing. It appears to be stuck in "Networking" which is confusing since it is not a networking problem.

Once a failure like this is diagnosed, it is still not straightforward or easy to fix.

This PR adds the 'nobootwait' or 'nofail' option to the `/etc/fstab` line of the mounted disk. This allows instances to boot even if the volume is missing. 'nofail' is used on all instances except Ubuntu 14 instances which use 'nobootwait'. These fstab changes are made for ephemeral disks and mounted volumes.

### DIFF
--- a/ansible/roles/atmo-ephemeral-mount/defaults/main.yml
+++ b/ansible/roles/atmo-ephemeral-mount/defaults/main.yml
@@ -4,3 +4,4 @@ disk_mounted_elsewhere: false
 
 MOUNT_PATH: '/scratch'
 MOUNT_TYPE: 'ext4'
+MOUNT_OPTS: "nofail"
--- a/ansible/roles/atmo-ephemeral-mount/defaults/main.yml
+++ b/ansible/roles/atmo-ephemeral-mount/defaults/main.yml
@@ -4,4 +4,5 @@ disk_mounted_elsewhere: false
 
 MOUNT_PATH: '/scratch'
 MOUNT_TYPE: 'ext4'
-MOUNT_OPTS: "nofail"
+MOUNT_OPTS: 'default,nofail'
+MOUNT_PASSNO: '2'
--- a/ansible/roles/atmo-ephemeral-mount/defaults/main.yml
+++ b/ansible/roles/atmo-ephemeral-mount/defaults/main.yml
@@ -4,5 +4,5 @@ disk_mounted_elsewhere: false
 
 MOUNT_PATH: '/scratch'
 MOUNT_TYPE: 'ext4'
-MOUNT_OPTS: 'default,nofail'
+MOUNT_OPTS: 'defaults,nofail'
 MOUNT_PASSNO: '2'
--- a/ansible/roles/atmo-ephemeral-mount/tasks/main.yml
+++ b/ansible/roles/atmo-ephemeral-mount/tasks/main.yml
@@ -1,9 +1,10 @@
 ---
 
-- name: Set MOUNT_OPTS Ubuntu < 16
-  set_fact:
-    MOUNT_OPTS: 'defaults,nobootwait'
-  when: ansible_distribution == "Ubuntu" and ansible_distribution_major_version < "16"
+- name: Gather OS specific variables
+  include_vars: "{{ item }}"
+  with_first_found:
+    - "{{ ansible_distribution }}-{{ ansible_distribution_major_version}}.yml"
+    - "{{ ansible_distribution }}.yml"
 
 - name: 'See if we have an ephemeral disk'
   stat:
--- a/ansible/roles/atmo-ephemeral-mount/tasks/main.yml
+++ b/ansible/roles/atmo-ephemeral-mount/tasks/main.yml
@@ -33,6 +33,8 @@
       src: '{{ ephem_stat.stat.lnk_source }}'
       fstype: '{{ MOUNT_TYPE }}'
       state: 'mounted'
+      opts: 'nobootwait'
+      passno: '2'
 
   - name: 'Data loss warning for Ephemeral storage'
     copy:
--- a/ansible/roles/atmo-ephemeral-mount/tasks/main.yml
+++ b/ansible/roles/atmo-ephemeral-mount/tasks/main.yml
@@ -2,7 +2,7 @@
 
 - name: Set MOUNT_OPTS Ubuntu < 16
   set_fact:
-    MOUNT_OPTS: 'nobootwait'
+    MOUNT_OPTS: 'default,nobootwait'
   when: ansible_distribution == "Ubuntu" and ansible_distribution_major_version < "16"
 
 - name: 'See if we have an ephemeral disk'
@@ -39,7 +39,7 @@
       fstype: '{{ MOUNT_TYPE }}'
       state: 'mounted'
       opts: '{{ MOUNT_OPTS }}'
-      passno: '2'
+      passno: '{{ MOUNT_PASSNO }}'
 
   - name: 'Data loss warning for Ephemeral storage'
     copy:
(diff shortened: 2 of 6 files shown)


calvinmclean:
Closing in favor of #139

ansible/roles/atmo-mount-volume/tasks/main.yml
```diff
@@ -1,5 +1,10 @@
 ---
 
+- name: Set MOUNT_OPTS Ubuntu < 16
+  set_fact:
+    MOUNT_OPTS: 'defaults,nobootwait'
+  when: ansible_distribution == "Ubuntu" and ansible_distribution_major_version < "16"
```
cdosborn: If a user passes MOUNT_OPTS they'll be ignored for these defaults if on ubuntu 16

calvinmclean: Ah yeah, sloppy merge on my part. Need to change it to match edwin's vars. I need to double-check if changing to `nobootwait` is even necessary

cdosborn: No sweat, merging was nice. There are ways in git to create merge that is only symbolic, meaning the merge shows two ancestors but actually ignores all changes introduced in one of them. I probably would have done that here. but doesn't really matter.

ansible/roles/atmo-ephemeral-mount/tasks/main.yml
```diff
@@ -1,5 +1,10 @@
 ---
 
+- name: Set MOUNT_OPTS Ubuntu < 16
+  set_fact:
+    MOUNT_OPTS: 'defaults,nobootwait'
+  when: ansible_distribution == "Ubuntu" and ansible_distribution_major_version < "16"
+
```
cdosborn: Rather than set_fact, wrap the code where its needed like so:
```
   - block:
       - name: 'Ephemeral disk mounted'
         mount:
           path: '{{ MOUNT_PATH }}'
           src: '{{ ephem_stat.stat.lnk_source }}'
           fstype: '{{ MOUNT_TYPE }}'
           state: 'mounted'
           opts: '{{ MOUNT_OPTS }}'
           passno: '{{ MOUNT_PASSNO }}'
     vars:
        UBUNTU_16: ansible_distribution == "Ubuntu" and ansible_distribution_major_version < "16",
        MOUNT_OPTS: 'defaults,{{ 'nobootwait' if UBUNTU_16 else 'nofail' }}'
```
Not sure if that'll work, but it completely avoids the disastrous global side effects of set_fact.

cdosborn: lol disastrous is a little extreme

edwins: don't use set fact for this purpose... i suggest using ansible defaults

edwins: @calvinmclean I just re-tested 14.04 with atmo-volume-mount with the same "defaults,nofail" and it appears to work -- including volume remounts upon reboot. Having a separate mount options for ubuntu 14 shouldn't be necessary. If you do, let me know if you discover otherwise.

calvinmclean: Interesting, I just finished testing also and I found that having `nofail` causes instances to fail a reboot. Here is how I tested:
- Launch Ubuntu 14 instance
- Attach volume through Atmosphere
- Verify that `/etc/fstab` shows `defaults,nofail` option
- Remove volume (through Atmosphere or OpenStack CLI)
- Verify that `/etc/fstab` is unchanged
- Hard reboot

and my Ubuntu 14 instances are stuck in Networking.

calvinmclean: I re-did my test with `nobootwait` and the Ubuntu 14 instances successfully reboot

ansible/roles/atmo-ephemeral-mount/defaults/main.yml
```diff
@@ -4,3 +4,5 @@ disk_mounted_elsewhere: false
 
 MOUNT_PATH: '/scratch'
 MOUNT_TYPE: 'ext4'
+MOUNT_OPTS: 'defaults,nofail'
```
cdosborn: Does this have any remaining effect now? Think it can be deleted.

ansible/roles/atmo-mount-volume/defaults/main.yml
```diff
@@ -1,3 +1,7 @@
 ---
 
 ATMO_USER_GROUP: "users"
+
+VOLUME_OPTS: "defaults,nofail"
```
cdosborn: Think this can be delted too

ansible/roles/atmo-mount-volume/tasks/main.yml
```diff
@@ -16,7 +16,12 @@
     src: '{{ VOLUME_DEVICE }}'
     state: 'mounted'
     path: '{{ VOLUME_MOUNT_LOCATION }}'
+    passno: '{{ VOLUME_PASSNO }}'
+    opts: '{{ VOLUME_OPTS }}'
   register: mount_result
+  vars:
+    UBUNTU_14: ansible_distribution == 'Ubuntu' and ansible_distribution_major_version < '16'
```
cdosborn: Sort an edge case, but you're really capturing anything less than ubuntu 16, like ubuntu 12. Would change to `UBUNTU_16: ansible_distribution == 'Ubuntu' and ansible_distribution_major_version == '16'` and update the ternary.

cdosborn: Actually i think my suggestion is bad. I'm okay with how it is,  the change for ubuntu 16 is probably going to extend into ubuntu 18 etc

calvinmclean:
For some reason, Ubuntu 14 instances still end up with the `nofail` option instead of `nobootwait`. Unfortunately, I have to get to class and will continue this on Friday.
It seems like `set_fact` might be the better option after all. I don't see what global side effects it could have since it sets the facts on a host-by-host basis.

edwins:
A couple of comments: 

1) while I like the creative use of vars attached to mount task, I think that approach actually is a little unintuitive and may be unclear to typical, non-developer folks who are writing ansible. The goal of atmosphere-ansible was that sites would fork, extend, and add additional functionality as they needed, it wasn't just for cyverse/jetstream only.

For better readability, I would prefer using ansible defaults (given their very low priority in evaluation) and then overriding them either using ansible vars (directory) or explicity setting them in a task (e.g. set_fact). Let's discuss this in person if there are other considerations about variable declaration.

@calvinmclean let's meet the next time you're in the office about the ubuntu 14 discrepancy. We can then compare our ubuntu 14 tests. The variable discussion may be moot (for now) if ubuntu 14 actually does accept nofail, as my testing is showing.

calvinmclean:
Here is an update of the current state of the branch:
- Use `defaults/main.yml` for `defaults,nofail`
- Use `vars/Ubuntu-14.yml` and `with_first_found` for `defaults,nobootwait`
- Tested by mounting a volume through Atmosphere, removing from OpenStack CLI (so that the entry remains in `/etc/fstab`), and rebooting instance
- Works on Ubuntu 14, Ubuntu 16, and CentOS

cdosborn:
Closing for squashed variant.
https://github.com/cyverse/atmosphere-ansible/pull/141

PR #137, opened 2018-02-21
Problem: XTerm is unusable and tasks that perform GUI actions are in inconsistent locations

calvinmclean:
### Background
This PR originates from PR #129 which intended to improve the font in XTerm because it is unusable by default. However, we decided that a better solution would be to install a different terminal (xfce4-terminal) because XTerm has more problems other than just the font.

The previous PR added the tasks to fix font in the `atmo-vnc` role, but I realized this doesn't quite fit. Also, tasks for changing desktop background were in that role, and tasks to install a web browser were in their own role, so I saw the opportunity to consolidate these tasks in a role `atmo-gui-tweaks`. 

### Changes
- Add tasks to install xfce4-terminal on instances with XFCE in `atmo-gui-tweaks`
- Add tasks to install gnome-terminal on instances with Gnome in `atmo-gui-tweaks`
- Move desktop background tasks and files from `atmo-vnc` to `atmo-gui-tweaks`
- Move web browser install from `install-browser` role to `atmo-gui-tweaks`

This PR has been tested launching featured images. GUI tweaks are still performed correctly and in the correct situations.


cdosborn:
Havn't had a change to take a look yet. Want to make sure that gui tweaks never prevent an instance from making it to active.

c-mart:
^ Agreed, I approve as long as this doesn't break launching. Sounds like Calvin has at least tested with all featured images. Most of the things in this role are non-critical and so *shouldn't* hold up instance deployment if they fail. So, maybe we should set `ignore errors: true` for all tasks in the `gui-tweaks` and `install-browser` roles.

calvinmclean:
Latest commit adds `ignore_errors` to all tasks. I tested it without errors and with errors (by changing file/package names to non-existent ones) and instances made it to active.

cdosborn:
Can you look into a way to achieve the desired behavior w/o calling ignore_errors on each task, it makes for a semi-useless role. I would prefer us to have a separate play, with a single ignore_errors, which imports all the gui roles. I'll look for something like that.

cdosborn:
Going to try something real quick. I'll report back.

PR #138, opened 2018-02-28
small fix to mount task to remove device entries in fstab

edwins:
Small change that will set the state to absent, which will cleanly remove the corresponding fstab entries


cdosborn:
Seems harmless to add. Why is it needed?

edwins:
without absent to the mount task, then the fstab entry remains. if the instance/snapshot/image is booted with the "orphaned" fstab entry in the past the instance would not boot. Along with the fixed fstab entry in the volume_mount, then it will still leave dirty fstab entries that may not have problems, but overtime may cause mount issues if there are conflicting entries.

cdosborn:
Sounds good

PR #139, opened 2018-02-28
fixed the passno and mount options for the mount task in atmo-mount-volume role

edwins:
Added a new default variable named VOLUME_PASSNO, which maps the the fstab option for passno, and defaults to 2 (i.e. non-root filesystems).


edwins:
added new variable named VOLUME_OPTS that allows for options to be passed to mount task. The default value is "default,nofail". The nofail will ensure that the system boots correctly when the disk does not exist, which may occur in edge cases of an instance going from inactive->active, a volume was detached external to atmosphere (e.g. an recovery situation) or an instance being imaged/snapshotted outside of Atmosphere.

cdosborn:
@calvinmclean  could you add your changes to this pr. This pr is also using "default" which i believe is a typo for "defaults"

cdosborn:
Closing in favor of: https://github.com/cyverse/atmosphere-ansible/pull/134

PR #140, opened 2018-02-28
 Create desktop shortcut to change the display resolution

calvinmclean:
The default display resolution for VNC sessions is 1024x768, which does not give the user much room to work with. If users browse through the wiki they would be able to find the command to change the resolution, but I assume many don't know that is an option. This PR will allow users to easily choose from all the available resolution options.

32623053-254ec418-c542-11e7-9000-e432b893fc10

re-created #130 because this moves everything to the new `atmo-gui-tweaks` role from #137 (which should be merged first)

### DIFF
--- a/ansible/roles/atmo-gui-tweaks/files/change_resolution.desktop
+++ b/ansible/roles/atmo-gui-tweaks/files/change_resolution.desktop
@@ -1,0 +1,8 @@
+[Desktop Entry]
+Version=1.0
+Name=change_desktop_resolution
+comment=Change desktop resolution
+Exec=/opt/cyverse_change_resolution.sh
+Terminal=true
+Type=Application
+Categories=Utility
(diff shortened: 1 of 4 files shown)


cdosborn:
I would prefer us not to be putting things on users' desktops.

julianpistorius:
I actually like the idea of putting this directly on the desktop. Make their lives as easy as possible. Otherwise they won't know it exists.

calvinmclean:
If it helps at all, the actual script is located in `/opt` like other cyverse tools and the thing on the desktop is just a "desktop entry"

cdosborn:
I'm okay with it going in. To explain where i'm coming from: I personally wouldn't want to buy a new pc and see a desktop full of icons (or any for that matter). It's even worse if the icon is just a shell script. If it were up to me i'd rather us just change the default resolution or something rather then give the user a bash script icon. I wouldn't click on a bash script that I didn't write.

calvinmclean:
I should be able to change it to a "menu entry" so it is in the application menu if that is preferable.
Also I'm open to the idea of just changing the default resolution.

calvinmclean:
Closing this in favor of a new PR (coming soon) that adds VNC Resolution to the settings page

PR #142, opened 2018-03-07
Add missing ATMO_USER var to script env

cdosborn:
Problem: DEPLOY_SCRIPTS are not called with ATMO_USER env variable
Solution: Use the shell module which supports setting variables

From the code this just looks like an oversight, because ASYNC_SCRIPTS get the
variable.


cdosborn:
This was tested during the v31 release as a hotfix.

Issue #126, opened 2017-10-25
When vms get a new IP address, web desktop breaks


cdosborn:
I helped triage the following ticket where a user had a desktop, but applications failed to launch. In fact I could not launch applications from the terminal.

I tried the following:
```
export DISPLAY=<can be :2 for vnc or :5 for guacamole>
export AUTHORITY=<path to Xauthority usually in ~/.Xauthority>
```
```
$ firefox 
No protocol specified
Failed to connect to Mir: Failed to connect to server socket: No such file or directory
Unable to init server: Could not connect: Connection refused
Error: cannot open display: :5
```
When I ran `xauth list` there were no cookies for the current vm's new ip/hostname. This was the issue. The cookies must match the current hostname. All I had to do was add new cookies. See this well written stackoverflow answer:
https://stackoverflow.com/questions/20611783/after-changing-hostname-gedit-and-other-x-clients-dont-open/20612084#20612084

PR #150, opened 2018-05-14
ATMO-2077: Fix VNC kill command so that new VNC servers can be started with correct xauth configuration

calvinmclean:
## Description

Problem: when instances are resumed, the existing VNC servers are not killed and restarted, so xauth is not updated with the new hostname. This would cause Web Desktop to refuse to open new windows.

The task that is supposed to kill all existing VNC servers using the command `vncserver -kill :$DISPLAY` does not succeed on instances being resumed. The changes here simply fix the kill task using `pkill` so that new VNC servers can be started. The new VNC servers will naturally update xauth.

More info can be found in Issue #126 and [ATMO-2077](https://pods.iplantcollaborative.org/jira/browse/ATMO-2077)

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,3 +31,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 - On CentOS, the dhclient.d script `ntp.sh` is configured to remove previously
   configured NTP servers in ntp.conf when installing DHCP-provided NTP
   servers. ([#146](https://github.com/cyverse/atmosphere-ansible/pull/146))
+
+### Fixed
+
+- Fixed task to kill VNC servers using `pkill` instead of `vncserver -kill :$DISPLAY` ([#150](https://github.com/cyverse/atmosphere-ansible/pull/150))
--- a/ansible/roles/atmo-vnc/tasks/realvnc.yml
+++ b/ansible/roles/atmo-vnc/tasks/realvnc.yml
@@ -106,9 +106,7 @@
     mode: 0755
 
 - name: kill all running vncserver sessions
-  shell: >
-    /bin/su '{{ ATMOUSERNAME }}' -c "{{ VNC_EXECUTABLE }} -kill :'{{ item }}'"
-  with_items: [1, 2, 3, 4, 5, 6, 7, 8]
+  command: 'pkill Xvnc-core'
   failed_when: False
 
 - name: Run VNC commands (Add license, set password, start server)


ansible/roles/atmo-vnc/tasks/main.yml
```diff
@@ -26,6 +26,10 @@
     has_gui: '{{ xsession.stat.exists and xterm.stat.exists }}'
 
 - block:
+  - include: xauth.yml
+    ignore_errors: yes
+    with_items: ['1', '2', '5']
+
```
cdosborn: Interesting approach for with_items + block.

ansible/roles/atmo-vnc/tasks/xauth.yml
```diff
@@ -0,0 +1,12 @@
...
+- block:
+    - name: Get xauth cookies
+      shell: "xauth list | grep :{{ item }} | awk '{print $3}'"
+      register: cookies
+
+    - name: Re-add cookies with new hostname
+      shell: 'xauth add "$(hostname)/unix:{{ item }}" MIT-MAGIC-COOKIE-1 {{ cookies.stdout_lines[0] }}'
+      when: cookies is defined and cookies.stdout_lines|length > 0
+  become: yes
+  become_user: '{{ ATMOUSERNAME }}'
```
cdosborn: If you look at the man page, just below "add" there is a "generate" command. I think the approach can be altered to: for each display, generate the cookie for that display.

calvinmclean: I squashed it all to one commit with a better message (in my opinion). I tried out using generate and could only get it to work if xauth was completely empty. So after generating for `:1`, it doesn't work for `:2` and `:5`. I'm looking to find out why this is happening but I don't see much

calvinmclean:
I completely changed this commit to properly kill old VNC servers, so starting new VNC servers as usual will create the correct xauth entries. I believe this is the correct fix for this bug because it fixes it at the cause of the problem instead of adding extra complexity.

cdosborn:
Can you test if we actually need to remove xauth list entries. We shouldn't have to worry about that. Then this change could just be fixing the killing of vnc sessions.

calvinmclean:
It is not needed, I just added it to keep stuff clean. I'll remove the task and update PR name and description

cdosborn:
When I pulled down the latest master and compared your tip:
```
$ git graph cyverse/master calvinmclean/ATMO-2077
* afd1fa0 (calvinmclean/ATMO-2077) Properly kill old VNC servers to avoid unrespons..
| *   ecf234a (cyverse/master) Merge pull request #146 from c-mart/jetstream-22
| |\
|/ /
| * 1dfd7ef (HEAD -> jetstream-22, c-mart/jetstream-22) Add changelog entry
| * b197550 Skip task to place ntp.sh if distro-specific tem..
| * ed5deab Fix command to renew DHCP lease on CentOS
| * 7fa07c3 Adding CentOS-specific ntp.sh files
| * 270d15e Configure CentOS to disregard previously configu..
|/
* 656c3e2 Merge pull request #145 from calvinmclean/ATMO-2..
```
You'll see your commit afd1fa0 was based on an old version of cyverse/master. I went ahead and rebased your work. You want to make sure to rebase your work, so that your testing your changes against the latest code base.

Take a look at the changelog and add an entry beneath the unreleased section in this pr. I'll then bring it in. Note that because I rewrote your history, you'll need to work off the latest branch on your remote. 

Here's the general process, lmk if any of that doesn't make sense.
```
git fetch <your remote> ATMO-2077
git checkout ATMO-2077
git reset --hard <your remote name>/ATMO-2077
# Create the changelog pr
git push <your remote> ATMO-2077
```

calvinmclean:
Done! Thanks for your help

PR #151, opened 2018-05-17
Use XFCE or Gnome commands to change background 

calvinmclean:
## Description

I noticed that our new CentOS 7 image did not get the correct desktop background change. This is because it uses XFCE desktop while the Ansible role expected CentOS instances to be using Gnome. This fix checks if XFCE commands are available on the instance, and if not, it uses Gnome commands instead.

This is a fix for CentOS 7, but it is also a big improvement to the task because it uses the correct commands to change the desktop background for XFCE instead of replacing the original background file with our own. It also gets rid of the specific checks for Ubuntu 14 and Ubuntu 16+.

### DIFF
--- a/ansible/roles/atmo-gui-tweaks/tasks/desktop-background.yml
+++ b/ansible/roles/atmo-gui-tweaks/tasks/desktop-background.yml
@@ -2,27 +2,32 @@
 - name: Copy desktop background to instance
   copy:
     src: "{{ background_img }}"
-    dest: /usr/share/backgrounds/cloud_background.jpg
+    dest: '{{ background_dest }}'
 
-- name: Change desktop background on CentOS
+- name: check if XFCE is desktop environment
+  stat:
+    path: '/usr/bin/xfce4-session'
+  register: xfce
+
+- name: desktop environment is XFCE
+  set_fact:
+    change_background_cmd: 'xfconf-query -c xfce4-desktop --property /backdrop/screen0/monitor0/workspace0/last-image --create --type string --set'
+  when: xfce.stat.exists
+
+- name: check if Gnome is desktop environment
+  stat:
+    path: '/usr/bin/gnome-session'
+  register: gnome
+
+- name: desktop environment is Gnome
+  set_fact:
+    change_background_cmd: 'gconftool-2 -s -t string /desktop/gnome/background/picture_filename'
+  when: gnome.stat.exists and not xfce.stat.exists
+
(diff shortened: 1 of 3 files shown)


ansible/roles/atmo-gui-tweaks/tasks/desktop-background.yml
```diff
@@ -2,27 +2,33 @@
...
+
+- name: desktop environment is Gnome
+  set_fact:
+    change_background_cmd: 'gconftool-2 -s -t string /desktop/gnome/background/picture_filename'
+  when: gnome.stat.exists and not xfce.stat.exists
+
+- name: Change desktop background
+  become: yes
+  become_user: '{{ ATMOUSERNAME }}'
+  shell: 'export $(dbus-launch); DISPLAY=:{{ item }} {{ change_background_cmd }} "{{ background_dest }}"'
```
cdosborn: I believe you can remove the `with_items` here. If you wouldn't mind throwing a comment in to say why dbus-launch is required. If DISPLAY must be set i think we can just use `:0`.

calvinmclean:
Works great without $DISPLAY. Rebased and pushed the changes. Thanks for your help and D-bus research!

calvinmclean:
Added a comment

ansible/roles/atmo-gui-tweaks/tasks/desktop-background.yml
```diff
@@ -2,27 +2,31 @@
...
+  register: gnome
+
+- name: desktop environment is Gnome
+  set_fact:
+    change_background_cmd: 'gconftool-2 -s -t string /desktop/gnome/background/picture_filename'
+  when: gnome.stat.exists and not xfce.stat.exists
+
+# dbus-launch is used here to launch the dbus-daemon and appropriate dekstop session and gather the environment variables that it creates.
+# dbus-launch is generally run as part of user login, but that has not happened yet so it is necessary here.
+- name: Change desktop background
```
cdosborn: Basically launching dbus is the wrong thing to do here-its definitely a hack. You need to explain not that we're launching dbus nor what dbus does, but that its a hack and **why** the hack is necessary. I realize that you may not have the holistic reason why its a hack.

We're creating a rogue dbus just so that gsettings will be satisfied. This is the wrong way to use gsettings, because an existing user session + windows would have its own dbus, and never see our updated setting. However, it's okay in this instance, because we nuke the users' desktop. The new desktop will get the updated settings from a file.

calvinmclean:
Finished testing latest changes

PR #152, opened 2018-05-25
Do not change permissions on all files a user's home directory

cdosborn:
## Description

Do not change permissions on all files in user home

Rather than changing permissions on every deploy, only change permissions if we create the user directory.

The original problem is that a user was timing out during deploy and it was hanging at this task. The user had 6mil files in their home. I don't think changing permissions would take more than 32min which is the upper limit for the user deploy task, but we shouldn't be changing permissions anyway.


cdosborn:
I tested this, and hotfixed in j7m.

PR #153, opened 2018-06-18
Change ownership of ephemeral mount to atmosphere user instead of root

calvinmclean:
## Description

The ephemeral disk mounted on `/scratch` is owned by root by default. This adds a task to change the ownership of the `/scratch` directory to the user that owns the instance.


ansible/roles/atmo-ephemeral-mount/tasks/main.yml
```diff
@@ -48,6 +48,14 @@
...
       src: 'DATA-LOSS-WARNING.txt'
       dest: '{{ MOUNT_PATH }}/DATA-LOSS-WARNING.txt'
+
+  - name: Change directory ownership
+    file:
+      state: directory
+      path: '{{ MOUNT_PATH }}'
+      recurse: yes
+      owner: '{{ ATMOUSERNAME }}'
+      group: "{{ ATMO_USER_GROUP }}"
```
cdosborn: We don't want to change permissions of every file in their scratch directory, this can take a long time and lock up the file system if they have many files. We should just not recurse here.

calvinmclean: FIxed

ansible/roles/atmo-ephemeral-mount/defaults/main.yml
```diff
@@ -2,6 +2,8 @@
 
 disk_mounted_elsewhere: false
 
+ATMO_USER_GROUP: "users"
+
```
cdosborn: Can you see if j7m uses the "users" group?

PR #154, opened 2018-06-29
ATMO-2149: Atmosphere deploy without Subspace

calvinmclean:
## Description

Replace Subspace by using Ansible's PlaybookCLI to deploy instances.

When the related branch of Atmosphere stopped using Subspace, we needed a new place to store our custom Callback plugin for logging so this is its new home. The new Callback plugin is similar to Ansible's default stdout but has the banners removed. In addition, it provides the instance UUID and username of the owner.

Logs are now organized in directories based off username and filenames of the instance UUID: `/opt/dev/atmosphere/logs/atmosphere_deploy.d/<username>/<UUID>.log`.

~~Atmosphere-Ansible is updated to be compatible with Ansible 2.5 free of deprecation warnings.~~

~~The new method of running playbooks in Atmosphere caused a lot of issues with unicode strings that had to be converted to regular strings. This caused an error with deploy scripts because they would be copied to a file on the instance, but the newline characters (`\n`) would not be evaluated to newlines and the scripts would fail. To fix this, I split the script string at the newlines into a list, and write it to a template line-by-line.~~

When creating instance deploy scripts, they would be saved in a dictionary that had to be passed to Ansible as a string, so the strings in the dictionary would be changed to the literal representation (replacing '\n' with '\\n') so instead it is passed as a list of lines and joined using jinja2.

Related PRs:
  - [Atmosphere PR #631](https://github.com/cyverse/atmosphere/pull/631)
  - [Clank PR #262](https://github.com/cyverse/clank/pull/262)


ansible/plugins/callback/atmo_logger.py
```diff
@@ -0,0 +1,274 @@
...
+        if 'exception' in result._result:
+            # Extract the error message and log it
+            error = result._result['exception'].strip().split('\n')[-1]
+            self.log(error)
+
+            # Remove the exception from the result so it's not shown every time
+            del result._result['exception']
+
+        # Else log the reason for the failure
+        if result._task.loop and 'results' in result._result:
```
cdosborn: This is fragile is it necessary?

ansible/plugins/callback/atmo_logger.py
```diff
@@ -0,0 +1,274 @@
...
+            msg = '  ok'
+
+        if delegated_vars:
+            msg += ": [%s -> %s]" % (result._host.get_name(), delegated_vars['ansible_host'])
+        else:
+            msg += ": [%s]" % result._host.get_name()
+
+        msg += " => (item=%s)" % (self._get_item(result._result),)
+
+        if (self._display.verbosity > 0 or '_ansible_verbose_always' in result._result) and not '_ansible_verbose_override' in result._result:
```
cdosborn: Fragile is it necessary?

ansible/plugins/callback/atmo_logger.py
```diff
@@ -0,0 +1,274 @@
...
+            self.log("  fatal: [%s]: UNREACHABLE! => %s" % (result._host.get_name(), self._dump_results(result._result)))
+
+    def v2_runner_on_no_hosts(self, task):
+        self.log("  skipping: no hosts matched")
+
+    def v2_playbook_on_task_start(self, task, is_conditional):
+        self.log("TASK [%s]" % task.get_name().strip())
+
+    def v2_playbook_on_play_start(self, play):
+        name = play.get_name().strip()
```
cdosborn: Does `get_name` guarantee a string or is it `None` sometimes?

ansible/roles/atmo-common/tasks/main.yml
```diff
@@ -36,7 +37,8 @@
     - timezone
 
 - name: update timezone manually
-  shell: /bin/su {{ item }} -c export TZ={{ TIMEZONE }}
+  become_user: '{{ item }}'
```
cdosborn: fwiw, I don't believe this shell command has any effect. `export` only affects the current and child processes.

ansible/roles/atmo-ephemeral-mount/tasks/main.yml
```diff
@@ -20,7 +20,7 @@
   when:
     - 'ephem_stat.stat.islnk is defined'
     - 'ephem_stat.stat.islnk == true'
-    - 'item.device == "{{ ephem_stat.stat.lnk_source }}"'
+    - 'item.device == "ephem_stat.stat.lnk_source"'
```
cdosborn: In order for this to work you'll need to remove the quotes around `ephem_stat...`. I believe you're comparing the device to the literal string "ephem_stat...".

You can test this in isolation of the rest of the playbook by creating an ephem_stat variable that holds a sample dictionary.

ansible/roles/atmo-user-deployment-scripts/tasks/main.yml
```diff
@@ -15,8 +15,8 @@
 
 
 - name: deploy_script_2a - copy scripts to available scripts directory
-  copy:
-    content: "{{ item.text }}"
+  template:
+    src: "script.j2"
```
cdosborn: Please figure out why the content isn't represented appropriately. My hunch is that `item.text` is a list of lines, if that is the case `{{ item.text | join('\n') }}` should work. But it should not be a list, did this issue crop up because of ansible 2.5?

calvinmclean: `item.text` was a str and would be copied into the file as a single line with literal `\n` characters in it:
```
#!/bin/bash\necho "wow" > /etc/wow.txt\necho "wow2" >> /etc/wow.txt
```

I made it into a list of lines using `splitlines()` in order to make this template approach work. However, your idea of using `{{ item.text | join('\n') }}` works and is better than using the template (I just tested it).

I did some tests using ansible 2.3 and 2.5 with a simple test playbook, and both copied content as expected so it is not a 2.5 issue. So it must be something about how the string is represented.

I believe it has something to do with [this method](https://github.com/cyverse/atmosphere/blob/37601a3397ee1f3bcadfe964ab7483abec8e82ea/core/models/boot_script.py#L74-L86) in `boot_script.py` so I will dig deeper there.

calvinmclean: In ansible, the `item.text` variable has `\\n` for newlines instead of `\n` which is why they are not evaluated. Just need to figure out why this happens so I need to learn more about the difference between str, unicode, and ascii

cdosborn: Check if the boot script model is storing the literal newline character, i'm guessing it is.That code shouldn't have an effect newlines are in the ascii range.

calvinmclean:
@cdosborn I was able to create a new version of the logging plugin. Instead of basically copying the stdout plugin and replacing `self._display.display(msg)` with `self.log(msg)`, I added a new class, `FakeDisplay`, and overrode the `display(msg)` method with the body from my `log(msg)` method. I then extended the default stdout plugin and only overrode `v2_playbook_on_play_start` in order to get the username and UUID info. I created a gist for it instead of pushing here in case it is too much of a hacky solution: https://gist.github.com/calvinmclean/5d68b4dd39c66d387ad3add5901a4267

However, it is missing the runtime stat so the play recap looks like:
```
2018-07-09 15:13:59 | 7db4b231-1051-4160-b17f-5d72d26a7ca1 | calvinmclean | PLAY RECAP
2018-07-09 15:13:59 | 7db4b231-1051-4160-b17f-5d72d26a7ca1 | calvinmclean | vm142-92.cyverse.org       : ok=37   changed=16   unreachable=0    failed=1
```

instead of:
```
2018-06-15 14:48:39 | ee67cb9d-3e14-4c65-a93e-bebea2f8bf0e | PLAY RECAP [vm142-19.cyverse.org] : ok: 4 changed: 1 unreachable: 0 skipped: 4 failed: 0 runtime: 0.6 seconds
```

I can override another method to add the runtime info

Note that it still uses `time` instead of `datetime`. I'm working on that next.

calvinmclean:
I decided to override the `v2_playbook_on_stats` method also because I preferred the one-line log and most of the default method could be removed so it didn't add too much complexity. The gist is updated

ansible/roles/atmo-user-deployment-scripts/tasks/main.yml
calvinmclean: So it looks like the type of the script text is `str` in atmosphere, and it prints out in the logs just as expected. In the `execute_playbooks()` method, after creating the list of string args (including extra_vars), the script text has `\\n` instead of `\n`, as if using the `repr()` method. This was due to the way I created the list of strings: `--extra_vars=%s" % extra_vars`

ansible/roles/atmo-common/tasks/main.yml
calvinmclean: Should I just delete this task?

calvinmclean:
check out my sweet rebase skillz. Remove Ansible 2.5 changes

ansible/plugins/callback/atmo_logger.py.j2
```diff
@@ -0,0 +1,98 @@
...
+    short_description: write playbook output to per-instance logs
+    version_added: 2.5
+    description:
+        - This plugin is modeled after the default playbook stdout (without the banners)
+        - It writes logs per-instance to `<log_path>/<username>/<instance_uuid>`
+    extends_documentation_fragment:
+      - default_callback
+    requirements:
+      - Whitelist in configuration
+      - `log_path` defined as a directory in `ansible.cfg` and writeable by the user running Ansible
```
cdosborn: This is not a requirement anymore right?

ansible/plugins/callback/atmo_logger.py.j2
```diff
@@ -0,0 +1,98 @@
...
+'''
+
+from ansible.plugins.callback.default import CallbackModule as CallbackModule_default
+from ansible.module_utils._text import to_text
+from ansible import constants as C
```
cdosborn: Move all imports above `DOCUMENTATION`. Convention is 3rd part libraries first, a blank line then all our libraries (int this case nothing),.

ansible/plugins/callback/atmo_logger.py.j2
```diff
@@ -0,0 +1,98 @@
...
+
+class DisplayLogger:
```
cdosborn: Make your display object inherit from Display in `lib/ansible/utils/display.py'

ansible/plugins/callback/atmo_logger.py.j2
```diff
@@ -0,0 +1,98 @@
...
+    overridden:
+        v2_playbook_on_play_start: only change is to get the 'username' and 'instance_id' from
+                                   'extra_vars' and save as an attribute
+        v2_playbook_on_stats: changed to add a runtime stat and print on one line
+    """
+    CALLBACK_VERSION = 2.0
+    CALLBACK_NEEDS_WHITELIST = True
+    CALLBACK_TYPE = 'notification'
+    CALLBACK_NAME = 'atmo_logger'
+    username = None
```
cdosborn: I believe this is unused

ansible/plugins/callback/atmo_logger.py.j2
```diff
@@ -0,0 +1,98 @@
...
+
+    def __init__(self):
+        super(CallbackModule, self).__init__()
+        self._display = DisplayLogger()
+        # Start counting time from creation to completion of exection.
+        self.start_time = datetime.now()
```
cdosborn: Should `self.start_time` be set in `on_play_start`?

ansible/roles/atmo-ntp/tasks/main.yml
```diff
@@ -113,7 +113,10 @@
   when: 'ansible_distribution == "CentOS"'
 
 - name: 'Renew DHCP lease for CentOS'
-  command: 'dhclient'
+  command: '{{ item }}'
+  with_items:
+    - 'pkill dhclient'
+    - 'dhclient'
```
cdosborn: This needs to be moved into a separate pr

ansible/roles/atmo-ssh-setup/tasks/ssh-setup.yml
```diff
@@ -4,11 +4,10 @@
...
 - name: Gather OS-specific variables
-  include_vars: "{{ item.var }}"
-  when: "{{ item.condition }}"
-  with_items:
-    - { var: 'Ubuntu.yml', condition: ansible_distribution == "Ubuntu" }
-    - { var: 'CentOS.yml', condition: ansible_distribution == "CentOS" }
+  include_vars: "{{ item }}"
+  with_first_found:
+    - "{{ ansible_distribution }}-{{ ansible_distribution_major_version}}.yml"
+    - "{{ ansible_distribution }}.yml"
```
cdosborn: This needs to be moved into a separate pr

ansible/ansible.cfg.j2
```diff
@@ -58,7 +58,7 @@ timeout = {{ ANSIBLE_SSH_TIMEOUT }}
 
 # logging is off by default unless this path is defined
 # if so defined, consider logrotate
-log_path = {{ ATMOSPHERE_ANSIBLE_LOG_DIR }}/ansible.log
+log_path = {{ ATMOSPHERE_ANSIBLE_LOG_DIR }}/atmosphere_deploy.log
```
cdosborn: Will this overwrite `logs/atmosphere_deploy.log`?

ansible/plugins/callback/atmo_logger.py.j2
```diff
@@ -0,0 +1,93 @@
+DOCUMENTATION = '''
```
cdosborn: At the top of this file you should mention that it is templated.

ansible/plugins/callback/atmo_logger.py.j2
```diff
@@ -0,0 +1,93 @@
...
+    callback: atmo_logger
+    type: notification
+    short_description: write playbook output to per-instance logs
+    description:
+        - This plugin is modeled after the default playbook stdout (without the banners)
+        - It writes logs per-instance to `<log_path>/<username>/<instance_uuid>`
```
cdosborn: You can template `<log_path>` here

CHANGELOG.md
```diff
@@ -31,6 +31,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 - On CentOS, the dhclient.d script `ntp.sh` is configured to remove previously
   configured NTP servers in ntp.conf when installing DHCP-provided NTP
   servers. ([#146](https://github.com/cyverse/atmosphere-ansible/pull/146))
+- New logging plugin that will create logs for each instance organized in username directories. The plugin is a template that has the log path added by the `./configure` script.
```
cdosborn: Link the pr in this entry

ansible/plugins/callback/atmo_logger.py.j2
cdosborn: I wonder if the template module automatically can do this with the configure ansible template message...

calvinmclean: This file is templated using the `./configure` script, not the ansible template module. I can change it to use the ansible template module and add `{{ ansible_managed }}` to the top of the file because that is a variable defined by default in `ansible.cfg` to show that a file is templated.

I used the configure script to keep it consistent with the `ansible.cfg`. Maybe we could make this change in a separate PR? It would require a few modifications to Clank, possibly a new role

calvinmclean: I made the change but the diff is still showing 🤔

cdosborn: I see it

PR #155, opened 2018-07-13
Upgrade Ansible version to the latest (2.6.1)

calvinmclean:
## Description

In order for Atmosphere-Ansible to be compatible with the latest Ansible (2.6.1), quite a few changes were required.

Deprecation fixes in `ansible.cfg`:
- `hostfile` changed to `inventory`
- instances of `sudo` changed to `become` (`sudo_user`, `sudo_exe`)
- `module_lang` deprecated

Deprecation fixes in roles:
- uses of `{{ var }}` in `when:` statements had to be changed to just `var`
- in packaging modules, `state: present` is used instead of `state: installed`
- replace instances of `system` or `systemctl` commands in shell modules with `service` module
- conditional tests using Jinja filters like `result|success` replaced with `result is success`
- specify `follow=yes` in template module that writes to a symlink because in Ansible 2.4 the default changed to `no`.

Bugs:
- in `atmo-kanki-irodsclient` role, the `apt` module was used with ftp URLs to download and install `.deb` files. This task began to fail with newer versions of Ansible, and with further investigation it looked like the download was succeeding because the error message would include `OK (12345 bytes downloaded)` where the number of bytes was the exact size of the file, so it is due to a bug in Ansible. A [recent commit](https://github.com/ansible/ansible/commit/86a63cf) added a check for http status codes to see if the download succeeded, but the status was `None` in this case, probably due to the fact that it is ftp now https (you can compare the check in that commit with the similar one used in the `get_url` module [here](https://github.com/ansible/ansible/blob/8606fb33f0e8954b588eaecec2d99b4a120fd4ad/lib/ansible/modules/net_tools/basics/get_url.py#L332)). I deal with this bug by downloading the files first, then installing. Maybe I should submit a bug fix to Ansible 🤔 

Related PR: Atmosphere [PR #635](https://github.com/cyverse/atmosphere/pull/635)

### DIFF
--- a/ansible/roles/atmo-kanki-irodsclient/vars/Ubuntu.yml
+++ b/ansible/roles/atmo-kanki-irodsclient/vars/Ubuntu.yml
@@ -12,6 +12,6 @@ dependencies:
   - g++
 
 irods_files:
-  - { url: 'ftp://ftp.renci.org/pub/irods/releases/4.1.9/ubuntu14/irods-runtime-4.1.9-ubuntu14-x86_64.deb', dest: '/tmp/irods-runtime.deb' }
-  - { url: 'ftp://ftp.renci.org/pub/irods/releases/4.1.9/ubuntu14/irods-icommands-4.1.9-ubuntu14-x86_64.deb', dest: '/tmp/irods-icommands.deb' }
-  - { url: 'ftp://ftp.renci.org/pub/irods/releases/4.1.9/ubuntu14/irods-dev-4.1.9-ubuntu14-x86_64.deb', dest: '/tmp/irods-dev.deb' }
+  - { url: 'ftp://ftp.renci.org/pub/irods/releases/4.1.9/ubuntu14/irods-runtime-4.1.9-ubuntu14-x86_64.deb', dest: '/tmp/irods-runtime.deb', checksum: '0c44062d184150469bf0662ee7aca888796067bc' }
+  - { url: 'ftp://ftp.renci.org/pub/irods/releases/4.1.9/ubuntu14/irods-icommands-4.1.9-ubuntu14-x86_64.deb', dest: '/tmp/irods-icommands.deb', checksum: '5e10ccc925c7c9de8a48d9e5575e67a2751de7d7' }
+  - { url: 'ftp://ftp.renci.org/pub/irods/releases/4.1.9/ubuntu14/irods-dev-4.1.9-ubuntu14-x86_64.deb', dest: '/tmp/irods-dev.deb', checksum: '6a46e71bc08c31bfb89ce98ec97b80f874e2a13e' }
--- a/ansible/roles/atmo-kanki-irodsclient/vars/Ubuntu.yml
+++ b/ansible/roles/atmo-kanki-irodsclient/vars/Ubuntu.yml
@@ -12,6 +12,6 @@ dependencies:
   - g++
 
 irods_files:
-  - ftp://ftp.renci.org/pub/irods/releases/4.1.9/ubuntu14/irods-runtime-4.1.9-ubuntu14-x86_64.deb
-  - ftp://ftp.renci.org/pub/irods/releases/4.1.9/ubuntu14/irods-icommands-4.1.9-ubuntu14-x86_64.deb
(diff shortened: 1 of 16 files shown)


ansible/roles/atmo-kanki-irodsclient/vars/Ubuntu.yml
```diff
@@ -12,6 +12,6 @@ dependencies:
   - g++
 
 irods_files:
-  - ftp://ftp.renci.org/pub/irods/releases/4.1.9/ubuntu14/irods-runtime-4.1.9-ubuntu14-x86_64.deb
-  - ftp://ftp.renci.org/pub/irods/releases/4.1.9/ubuntu14/irods-icommands-4.1.9-ubuntu14-x86_64.deb
-  - ftp://ftp.renci.org/pub/irods/releases/4.1.9/ubuntu14/irods-dev-4.1.9-ubuntu14-x86_64.deb
+  - { url: 'ftp://ftp.renci.org/pub/irods/releases/4.1.9/ubuntu14/irods-runtime-4.1.9-ubuntu14-x86_64.deb', dest: '/tmp/irods-runtime.deb' }
```
c-mart: Is this FTP download not protected with TLS? If it's not, then this represents a security issue, because we aren't checking that a man in the middle isn't serving us a different (possibly malicious) .DEB package.

[get_url](https://docs.ansible.com/ansible/latest/modules/get_url_module.html?highlight=get_url) has a `checksum` parameter. I think we should hard-code the checksums in this dictionary like this, and verify them in the download task:

```
{ url: 'ftp://ftp.renci.org/pub/irods/releases/4.1.9/ubuntu14/irods-icommands-4.1.9-ubuntu14-x86_64.deb', dest: '/tmp/irods-icommands.deb', sha1sum: 'deadbeefdeadbeefdeadbeefdeadbeefdeadbeef' }
```

(Incidentally, the iRODS consortium [doesn't serve their checksums via TLS](ftp://ftp.renci.org/pub/irods/releases/4.1.11/checksums.txt), which itself is an issue.)

calvinmclean: Good idea. Did you get that shasum by downloading and checking it yourself or did you find one posted somewhere?

calvinmclean: Never mind, I found them!

c-mart: Don't let that hold up this PR, I'll submit another to fix when this is merged.

ansible/roles/atmo-ntp/tasks/main.yml
```diff
@@ -52,15 +53,17 @@
     - time
 
 - name: install ntp prep package(s) -- Ubuntu only
-  shell: >
-    apt-get -qy install {{ NTP_PREP_PACKAGE.name }}
+  apt:
+    name: "{{ NTP_PREP_PACKAGE.name }}"
+    state: present
```
cdosborn: This ignores `"{{ NTP_PREP_PACKAGE.state }}"`. 

Now that we are no longer passing custom args to apt-get can we unify this with the package decl on lines 64-66.

ansible/roles/ldap/tasks/main.yml
```diff
@@ -233,19 +238,19 @@
...
-  when: ansible_distribution == "Ubuntu" and ansible_distribution_major_version|int < 16
-
-- name: reload dbus again just to be safe for Ubuntu 16.04 and later
-  shell: systemctl reload dbus
-  when: ansible_distribution == "Ubuntu" and ansible_distribution_major_version|int >= 16
+- name: restart dbus again just to be safe for Ubuntu
+  service:
+    name: 'dbus'
+    state: 'reloaded'
+  when: ansible_distribution == "Ubuntu"
```
cdosborn: :+1:

ansible/roles/atmo-ntp/tasks/main.yml
calvinmclean: Yeah I think that should work. I'll test it out tomorrow. I noticed there are a lot of weird practices in this role so maybe I should make a separate PR to fix a few things not related to the Ansible upgrade

PR #156, opened 2018-08-07
Restart network service to renew DHCP lease

calvinmclean:
## Description

Problem: The task that runs dhclient for CentOS fails after the first successful attempt because the dhclient process is already running.
Instead of force killing the process with pkill, restarting the network service will properly stop and start dhclient.


edwins:
Has this been tested with both centos 6 and 7?

calvinmclean:
I tested on both and it succeeded

julianpistorius:
Can confirm it worked on CenOS 7.4.

c-mart:
My testing yesterday was invalid (was looking at leases file in `/var/lib/dhclient/`). Today I `systemctl restart network`ed while `tail -f`ing `/var/log/messages` and I confirm that the DHCP lease is renewed within a few seconds of restarting the service.

PR #162, opened 2018-08-14
Improve iRods iCommands install process

calvinmclean:
## Description

This PR improves the iRods iCommands install process by deleting the unused `atmo-irods` role and removing many unused tasks from the `irods-icommands` role. In the second role, there were a lot of tasks designed to work with CentOS 5, which is no longer necessary since it is not supported. Removing these tasks makes the role much simpler.

This PR is based off changes in #161 so that has to be merged first.

- [ ] New test(s) included to reproduce the bug/verify the feature
- [ ] Documentation created/updated (include links)
- [ ] New variables contributed to atmosphere/Clank


calvinmclean:
since #161 was merged I will rebase this

PR #168, opened 2018-10-03
Change iRods iCommands URLs back to custom CyVerse version

calvinmclean:
## Description

Problem: [This commit](https://github.com/cyverse/atmosphere-ansible/commit/0741cffeb156d66df26ad7b1243086c89e7fea65#diff-7f4e8e482785f360de90be212be73cb7) replaced iRods iCommands URLs with the ones provided by iRods, but CyVerse was using a custom version with improvements to iRods Fuse.

This PR simply reverts the changes and uses the custom URLs

- [ ] New test(s) included to reproduce the bug/verify the feature
- [ ] Documentation created/updated (include links)
- [ ] New variables contributed to atmosphere/Clank

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 -->
 
 ## [Unreleased](https://github.com/cyverse/atmosphere-ansible/compare/v34-1...HEAD) - YYYY-MM-DD
+### Fixed
+  - Fix slow iRods Fuse performance
+    ([#168](https://github.com/cyverse/atmosphere-ansible/pull/168))
+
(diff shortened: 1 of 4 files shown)


cdosborn:
Commit history could be more useful. What was the exact error surfaced? Your commit message is redundant with the actual changes in the source. Would prefer to see something like.
```
Change iRods iCommands URLs back to custom CyVerse version

Problem: icommand <command> throws <error>

Reverting commit <blah>, above error occurred when we erroneously swapped out CyVerse specific client
```

Commit history should assist someone wondering if <error> was fixed in a given release. 

The changelog entry also coudl be more helpful. It sounds like it belongs under the `### CHANGED` heading. would prefer something like "Fixed broken icommand <blah> throwing <error>", people can look into it the history or pr to get the background and cause of the issue.

edwins:
It would be good to note that we'll revert back to the renci provided packages at sometime in the future once the irods fuse optimizations are implemented, targted for v4.2

PR #177, opened 2020-08-04
Update realvnc.yml

jlf599:

### DIFF
--- a/ansible/roles/atmo-vnc/tasks/realvnc.yml
+++ b/ansible/roles/atmo-vnc/tasks/realvnc.yml
@@ -91,23 +91,14 @@
     mode: 0755
     owner: '{{ ATMOUSERNAME }}'
 
-# NOTE template xstartup for centos 7 was in jetstream ONLY and should be evaluated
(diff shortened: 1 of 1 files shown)


ansible/roles/atmo-vnc/tasks/realvnc.yml
```diff
@@ -100,6 +100,15 @@
     owner: "{{ ATMOUSERNAME }}"
   when: (ansible_distribution == "CentOS") and (ansible_distribution_major_version == '7')
```
edwins, line 101: This line can be modified to end with "...ansible_distribution_major_version|int >= 7" . If you want to make that change, I'll accept. or I can make the change if you'd like.

While minor, this would save an extra check in ansible.

The "name" of this task should also be edited to reflect centos 7 & 8.

jlf599:
If you want to make the change, please go ahead. I'm an ansible newbie...just trying to save some time, but apparently that didn't work. :)

jlf599:
Though I have it open right now so I can fix.

edwins:
Perfect! Ty

